PR #42, opened 2015-03-04
Removing invisible non ASCII characters

antoniogarrote:
Noticed this characters when testing rdfstore-js.

Breaking the library when running in the browser.


RubenVerborgh:
Thanks, merged as 10caff00a91a290eab77ce8e326fbc031f97b37f.

Note that it doesn't break the library though; only when served with an incorrect encoding type. (Should be UTF8.)

PR #43, opened 2015-03-08
Removing trailing commas

antoniogarrote:
Google's closure compiler is complaining about these trailing commas:

ERROR - Parse error. IE8 (and below) will parse trailing commas in array and object literals incorrectly. If you are targeting newer versions of JS, set the appropriate language_in option.

### DIFF
--- a/N3.js
+++ b/N3.js
@@ -10,7 +10,7 @@ var exports = module.exports = {
   Store:        require('./lib/N3Store'),
   StreamParser: require('./lib/N3StreamParser'),
   StreamWriter: require('./lib/N3StreamWriter'),
-  Util:         require('./lib/N3Util'),
+  Util:         require('./lib/N3Util')
 };
 
 // Load submodules on first access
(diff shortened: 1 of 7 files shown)


RubenVerborgh:
Thanks, @antoniogarrote. However, as you know, trailing commas are technically not illegal in JavaScript, it's just IE8 having trouble with them (and I don't think it's an issue for the cases in which this is done here). In any case, [only IE versions >= 9 are supported](https://github.com/RubenVerborgh/N3.js/blob/master/package.json#L52).

Therefore, I'd prefer to keep the code style as-is, as it allows to simply append future code where needed. Would you be fine with that, or are there other reasons to remove trailing commas?

antoniogarrote:
Hi Ruben, sorry about the 'illegal'.

I left a comment explaining the reason. 
I'm running Google's Closure compiler on your code and it will report an error if I'm targeting something older than IE8.

Ignore this, if you prefer to keep the commas, I'll deal with it in a different way on my side.

Thanks.

RubenVerborgh:
No worries, thanks for letting me know :smile:

It's just that, if I remember correctly from when I was making N3.js browser-compatible, there were other incompatibilities with IE8 anyway that could not be solved easily (without sacrificing speed or other things). Therefore, I'd recommend to start from IE9 onwards. (It's time the Web moves on, anyway!)

PR #53, opened 2015-09-09
Single quote literals

ekerkenaar:
It would be nice to have single quote support when extracting data in triple format (for both single and multi-line string literals). Although the parser supports this the N3Util lacks this ability and extracting data becomes difficult.

Therefore, I did some searching around to see if this should be supported and found this: http://www.w3.org/TR/turtle/#turtle-literals Then I made some minor changes and added some basic tests to see if this is feasible.

I'n not sure what you guys think so please feel free to comment on this solution.

### DIFF
--- a/lib/N3Util.js
+++ b/lib/N3Util.js
@@ -18,7 +18,7 @@ var N3Util = {
 
   // Tests whether the given entity (triple object) represents a literal in the N3 library
   isLiteral: function (entity) {
-    return entity && entity[0] === '"';
+    return entity && (entity[0] === '"' || entity[0] === "'");
   },
 
   // Tests whether the given entity (triple object) represents a blank node in the N3 library
@@ -28,7 +28,7 @@ var N3Util = {
 
   // Gets the string value of a literal in the N3 library
   getLiteralValue: function (literal) {
-    var match = /^"([^]*)"/.exec(literal);
+    var match = /^["']((?:.|\s)*)["']/.exec(literal);
     if (!match)
       throw new Error(literal + ' is not a literal');
     return match[1];
--- a/test/N3Parser-test.js
+++ b/test/N3Parser-test.js
@@ -1314,7 +1314,7 @@ function shouldParse(parser, input) {
     var results = [];
     parser.parse(input, function (error, triple) {
       expect(error).not.to.exist;
-      if (triple)       
+      if (triple)
         results.push(triple);
       else {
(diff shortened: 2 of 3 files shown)


RubenVerborgh:
Hi @ekerkenaar,

Thanks for taking the time to work on this. However, I'm afraid there is some confusion between the Turtle syntax and the N3.js in-memory representation of literals. I'll try to explain it in-depth here.

Like other libraries, N3.js has an **in-memory representation** of entities (which are either resources with an IRI, or literals). While typical libraries in other programming languages such as Java would use a class, **N3.js uses regular strings to represent IRIs or literals". N3.js does this by giving special meaning to otherwise normal looking strings. The goal for this representation is a) to be fast b) to be unique, meaning that each unique N3.js representation should correspond to exactly one IRI or literal. (Note how this differs from Turtle, where IRIs and literals can be expressed in different ways.)

The N3.js representation follows the following rules (as explained in the [README](https://github.com/RubenVerborgh/N3.js/blob/master/README.md#triple-representation)):

- URLs, URIs and IRIs are simple strings
- Literals are represented as double quoted strings
  - For literals with a language or type, add a marker (`@` or `^^`) and the corresponding value as-is

The purpose of `N3.Util` is offering helper functions to work with this representation. So functions like `isLiteral` and `getLiteralValue` do not take Turtle syntax as an argument, but N3.js representations. By definition, such N3.js representations always use a single pair of double quotes—otherwise, they would not be unique. Also note that no escaping is performed. So the string of characters _ab"cd_ is represented in N3.js as `var object = '"ab"cd"'` (even though Turtle representations could be `"ab\"cd"`, `"""ab"cd"""`, `'ab"cd'` or `'ab\"cd'`).

Hence, the purpose of `isLiteral` and other methods is not to parse Turtle, because that is the purpose of `N3.Parser`. In fact, `N3.Parser` is the only class that takes Turtle as input; all other components of N3.js take N3.js memory representations as inputs. As such, `N3.Util` methods do not accept single-quoted literals—and this is by design, because single-quoted literals are Turtle syntax, not N3.js in-memory representations.

I hope this clarifies the current implementation. If you have any suggestion on how to make this more clear in the documentation, don't hesitate to discuss.

RubenVerborgh:
Closing this for now; feel free to reopen if needed.

ekerkenaar:
Thank you Ruben for taking the time to write this explanation. It is very clear.
The issue can indeed be closed. No need to reopen.

PR #57, opened 2015-10-14
Change interpretation of the graph parameter in the N3Store

rubensworks:
Setting the `graph` parameter to a falsy value will now, just like with the other parameters, be interpreted as a wildcard.
For fetching triples within the default graph, this graph explicitly needs to be provided as parameter value.
This default graph can be found in the store in the read-only field `defaultGraph`. This can be modified by providing a defaultGraph field in the constructor of the N3Store.

I'm aware that this will break existing implementations using the N3Store, in the sense that fetches from the store without a provided graph will now return all results instead of just the default graph. 
But I think this is for the better, to also streamline this with how this will happen in QPF.

### DIFF
--- a/test/N3Store-test.js
+++ b/test/N3Store-test.js
@@ -50,6 +50,23 @@ describe('N3Store', function () {
     });
   });
 
+  describe('An N3Store without a configured default graph', function () {
+    var store = new N3Store();
+
+    it('should have a dummy default graph', function () {
+      store.defaultGraph.should.eql('http://example.org/#defaultGraph');
+    });
+  });
+
+  describe('An N3Store with a configured default graph', function () {
+    var dg = 'http://example.org/#defaultGraph';
+    var store = new N3Store({ defaultGraph: dg });
+
+    it('should return that configured default graph', function () {
+      store.defaultGraph.should.eql(dg);
+    });
+  });
+
   describe('An N3Store with initialized with 3 elements', function () {
     var store = new N3Store([
       { subject: 's1', predicate: 'p1', object: 'o1' },
@@ -63,7 +80,7 @@ describe('N3Store', function () {
   });
 
   describe('An N3Store with 5 elements', function () {
-    var store = new N3Store();
+    var store = new N3Store({ defaultGraph: 'http://example.org/#defaultGraph' });
     store.addTriple('s1', 'p1', 'o1');
     store.addTriple({ subject: 's1', predicate: 'p1', object: 'o2' });
     store.addTriples([
@@ -77,15 +94,22 @@ describe('N3Store', function () {
     });
 
     describe('when searched without parameters', function () {
-      it('should return all items in the default graph',
+      it('should return all items',
         shouldIncludeAll(store.find(),
-                         ['s1', 'p1', 'o1'], ['s1', 'p1', 'o2'], ['s1', 'p2', 'o2'], ['s2', 'p1', 'o1']));
+                         ['s1', 'p1', 'o1', store.defaultGraph],
+                         ['s1', 'p1', 'o2', store.defaultGraph],
+                         ['s1', 'p2', 'o2', store.defaultGraph],
+                         ['s2', 'p1', 'o1', store.defaultGraph],
+                         ['s1', 'p2', 'o3', 'c4']));
     });
 
     describe('when searched with an existing subject parameter', function () {
-      it('should return all items with this subject in the default graph',
+      it('should return all items with this subject in all graphs',
         shouldIncludeAll(store.find('s1', null, null),
-                         ['s1', 'p1', 'o1'], ['s1', 'p1', 'o2'], ['s1', 'p2', 'o2']));
+                         ['s1', 'p1', 'o1', store.defaultGraph],
+                         ['s1', 'p1', 'o2', store.defaultGraph],
+                         ['s1', 'p2', 'o2', store.defaultGraph],
+                         ['s1', 'p2', 'o3', 'c4']));
     });
 
     describe('when searched with a non-existing subject parameter', function () {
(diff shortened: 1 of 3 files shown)


RubenVerborgh:
Hi Ruben, thanks for this.

I like the proposal. I don't mind the backwards compatibility; we'll just release this as a new minor version then. (It's still 0.x anyway.)

I just:
- wonder whether we could use `''` as a _default_ default graph. (Only problem: `''` _is_ a falsy value).
- need to check performance

For the latter, I'm thinking of replacing the `reduce`-`concat` loop by a `for`-`push` loop that is then concatenated once at the end:

```
quads.length === 1 ? quads[0] : a.concat.apply([],a)`).

File: test/N3Store-test.js [hunk 1]
```diff
@@ -77,15 +94,22 @@ describe('N3Store', function () {
     });
 
     describe('when searched without parameters', function () {
-      it('should return all items in the default graph',
+      it('should return all items',
```
RubenVerborgh: Agree on this; it's more logical.

rubensworks:
Initially I did use `''` as the *default* default graph, but many of the test were giving unexpected errors, probably because of `''` indeed being a falsy value. I could look into this some more, but I think this would require some more significant changes in the current code, but I could be wrong.

I see there is a performance test `perf/N3Store-perf.js`, I'll do some comparisons with that. I'll also look into changing that `reduce`-`concant` loop.

RubenVerborgh:
+1 and don't worry about `''` for now, having an actual URL in their might be a better idea.

rubensworks:
Here is the performance test output run on my machine:

    N3Store performance test (original-run1)
    - Adding 16777216 triples: 15125ms
    * Memory usage: 941MB
    - Finding all 16777216 triples 196608 times: 16786ms

    N3Store performance test (original-run2)
    - Adding 16777216 triples: 15136ms
    * Memory usage: 804MB
    - Finding all 16777216 triples 196608 times: 12968ms

    N3Store performance test (patch1)
    - Adding 16777216 triples: 15905ms
    * Memory usage: 920MB
    - Finding all 16777216 triples 196608 times: 12743ms

    N3Store performance test (patch2)
    - Adding 16777216 triples: 15479ms
    * Memory usage: 939MB
    - Finding all 16777216 triples 196608 times: 12668ms

(`patch1` is before 1c879a7)

RubenVerborgh:
Good good; I still wonder whether `for (var graph in graphs)` is faster than `Object.keys(graphs).forEach`; it will be on certain Node versions and other JS engines.

The perf tests only consider the one-graph case though; we might need to extend them with a one-graph and a multi-graph case:

```
var i, j, k, l;
for (i = 0; i < dim; i++)
  for (j = 0; j < dim; j++)
    for (k = 0; k < dim; k++)
      store.addTriple(prefix + i, prefix + j, prefix + k);
for (i = 0; i < dim; i++)
  for (j = 0; j < dim; j++)
    for (k = 0; k < dim; k++)
            for (l = 0; l < dim; l++)
                store.addTriple(prefix + i, prefix + j, prefix + k, prefix + l);
```

rubensworks:
The only reason I chose for `Object.keys(graphs).forEach` in this case was so I could use `return` in the cases where the iteration should immediately stop. Otherwise I had to use `continue` or change the structure of the `if`-conditions, which would also be possible if you prefer that.

Ok, I will extend the performance test so it also tests with multiple graphs.

RubenVerborgh:
@rubensworks I follow your arguments. While I also prefer `return`, `continue` would be acceptable here if performance is better. (There used to be a non-negligible overhead of creating the function needed for `forEach`, so we should avoid this in performance-critical paths like here.)

rubensworks:
@RubenVerborgh Ok, no problem, I'll change it.

rubensworks:
As can be seen in the last commit, I extended the performance test with a multi-graph case.
Below you can find the results of this test, first executed on the original code and then executed on the new code contained in this PR.

    N3Store performance test (original)
    - Adding 421875 triples in the default graph: 361ms
    * Memory usage for triples: 63MB
    - Finding all 421875 triples in the default graph 16875 times: 105ms

    - Adding 31640625 quads: 41208ms
    * Memory usage for quads: 1045MB
    - Finding all 31640625 quads 1687500 times: 15428ms

    N3Store performance test (patch)
    - Adding 421875 triples in the default graph: 351ms
    * Memory usage for triples: 63MB
    - Finding all 421875 triples in the default graph 16875 times: 117ms

    - Adding 31640625 quads: 39894ms
    * Memory usage for quads: 1039MB
    - Finding all 31640625 quads 1687500 times: 54108ms

The major difference here lies in the retrieval of the quads, which is a lot slower in the patched version.
This is mainly caused by the changed interpretation of the falsy graph value, this is also the reason why the perf test in the original code requires a little modification:

    for (i = 0; i < dim; i++)
      assert.equal(store.find(prefix + i, null, null, null).length, 0);
    for (j = 0; j < dim; j++)
      assert.equal(store.find(null, prefix + j, null, null).length, 0);
    for (k = 0; k < dim; k++)
      assert.equal(store.find(null, null, prefix + k, null).length, 0);
    for (l = 0; l < dim; l++)
      assert.equal(store.find(null, null, null, prefix + l).length, dimCubed);
      
In an attempt to make these interpretations more comparable, I changed the quads case to the following in both code versions:

    TEST = '- Finding all ' + dimTesseracted + ' quads ' + dimCubed * 4 + ' times';
    console.time(TEST);
    for (l = 0; l < dim; l++)
      assert.equal(store.find(null, null, null, prefix + l).length, dimCubed);
    console.timeEnd(TEST);

And these are the new results:

    N3Store performance test (original)
    - Adding 31640625 quads: 39917ms
    * Memory usage for quads: 1048MB
    - Finding all 31640625 quads 1687500 times: 15720ms

    N3Store performance test (patch)
    - Adding 31640625 quads: 39867ms
    * Memory usage for quads: 1046MB
    - Finding all 31640625 quads 1687500 times: 15215ms

So it looks like the performance is about the same.

RubenVerborgh:
Thanks; will test on different Node versions voor performance and merge.

RubenVerborgh:
Thanks, merged as c1f2906226e11c55cf639653abdfc42465d5e528.

PR #58, opened 2015-10-30
`addTriple` returns if the insertion changed the index

rubensworks:
It now returns whether or not the triple already existed in the store.
Small convenience feature which should not have any noticable impact, since the execution of `key2 in index2` is constant.
This could be made even faster if there was a function that allowed dictionary insertions while checking if the dictionary has changed, but don't think such a function exist.

I only added the `changed` check to the `subjects` index, I don't think it is possible for these indexes to differ? If so, I can add this check to the three indexes for that graph.

### DIFF
--- a/lib/N3Store.js
+++ b/lib/N3Store.js
@@ -55,12 +55,15 @@ N3Store.prototype = {
   // ## Private methods
 
   // ### `_addToIndex` adds a triple to a three-layered index.
+  // Returns if the index has changed, if the entry did not already exist.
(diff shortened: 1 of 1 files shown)


RubenVerborgh:
Thanks, merged as 64b34668a03f50899f80104e3606d9f5f6976a59.

RubenVerborgh:
Did the same for `removeTriple` in 04f4e21f4ccb351587dc00a3f26340b28d4bb10f.

PR #61, opened 2016-05-20
Improve N3Store lookup efficiency by maintaining an inverse index

rubensworks:
This is an improvement for the lookup-speed of the `N3Store`, as discussed in https://github.com/rubensworks/N3.js/commit/87cae2914751c93fdeb8c005b6c445d002b6f446.
Apparently, `Object.keys` became more expensive since one of the newer Node versions,
so this PR avoids calling this function every time on `this._entities` when doing lookups.
Instead, a inverse-index is maintained to avoid this operation.

I ran the performance tests to compare this new approach with the old one,
and it resulted in some interesting results:

```
2 variables for lookups (a.k.a original perf tests):
NEW:
    N3Store performance test
    - Adding 16777216 triples in the default graph: 18177.111ms
    * Memory usage for triples: 831MB
    - Finding all 16777216 triples to the default graph 196608 times: 6421.472ms

    - Adding 16777216 quads: 24892.776ms
    * Memory usage for quads: 676MB
    - Finding all 16777216 quads 1048576 times: 13141.350ms

OLD:
    N3Store performance test
    - Adding 16777216 triples in the default graph: 19465.032ms
    * Memory usage for triples: 808MB
    - Finding all 16777216 triples to the default graph 196608 times: 3206.058ms

    - Adding 16777216 quads: 24860.754ms
    * Memory usage for quads: 717MB
    - Finding all 16777216 quads 1048576 times: 19608.113ms
------------------------------------------------------------------------------------
1 variable for lookup:
NEW:
    - Finding all 16777216 triples to the default graph 65536 times: 1207.522ms
OLD:
    - Finding all 16777216 triples to the default graph 65536 times: 2133.597ms
------------------------------------------------------------------------------------
0 variables for lookup:
NEW:
    - Finding all 16777216 triples to the default graph 65536 times: 92602.621ms
OLD:
    - Finding all 16777216 triples to the default graph 65536 times: 283470.831ms
```

It appears that this new approach (precalculation of inverse index) is two times *slower* than the original approach for lookups with two variables (? ? o, ? p ?, s ? ?).
When increasing the selectivity of the lookups, this new approach becomes *faster* than the old one.
The new version is twice as fast as the old version for lookups with only one variables.
And it is three times faster than the old version for lookups without variables.

Like you have said before, the efficiency strongly depends on the access patterns.
So I'm not sure if you should accept this PR, since it makes lookups slower with two variables, but faster for the rest.
Maybe we should make two versions of the `N3Store` that are optimized for different cases.
Or maybe some more elaborate solution is possible.

### DIFF
--- a/lib/N3Store.js
+++ b/lib/N3Store.js
@@ -14,7 +14,8 @@ function N3Store(triples, options) {
   // `_entities` maps entities such as `http://xmlns.com/foaf/0.1/name` to numbers.
   // This saves memory, since only the numbers have to be stored in `_graphs`.
   this._entities = Object.create(null);
-  this._entities['><'] = 0; // Dummy entry, so the first actual key is non-zero
+  // `_entitiesInverse` is the inverse mapping of `_entities`, this is to avoid expensive inverse lookups.
+  this._entitiesInverse = Object.create(null);
   this._entityCount = 0;
   // `_blankNodeIndex` is the index of the last created blank node that was automatically named
   this._blankNodeIndex = 0;
@@ -88,7 +89,7 @@ N3Store.prototype = {
   // (for instance: _subject_, _predicate_, and _object_).
   // Finally, `graph` will be the graph of the created triples.
   _findInIndex: function (index0, key0, key1, key2, name0, name1, name2, graph) {
-    var results = [], entityKeys = Object.keys(this._entities), tmp, index1, index2;
+    var results = [], entityKeys = this._entitiesInverse, tmp, index1, index2;
 
     // If a key is specified, use only that part of index 0.
     if (key0) (tmp = index0, index0 = {})[key0] = tmp[key0];
@@ -169,9 +170,10 @@ N3Store.prototype = {
     // Instead, we have a separate index that maps entities to numbers,
     // which are then used as keys in the other indexes.
     var entities = this._entities;
-    subject   = entities[subject]   || (entities[subject]   = ++this._entityCount);
-    predicate = entities[predicate] || (entities[predicate] = ++this._entityCount);
-    object    = entities[object]    || (entities[object]    = ++this._entityCount);
+    var entitiesInverse = this._entitiesInverse;
+    subject   = entities[subject]   || (entities[entitiesInverse[++this._entityCount] = subject]   = this._entityCount);
+    predicate = entities[predicate] || (entities[entitiesInverse[++this._entityCount] = predicate] = this._entityCount);
+    object    = entities[object]    || (entities[entitiesInverse[++this._entityCount] = object]    = this._entityCount);
 
     var changed = this._addToIndex(graphItem.subjects,   subject,   predicate, object);
     this._addToIndex(graphItem.predicates, predicate, object,    subject);
--- a/lib/N3Store.js
+++ b/lib/N3Store.js
@@ -89,7 +89,9 @@ N3Store.prototype = {
   // (for instance: _subject_, _predicate_, and _object_).
   // Finally, `graph` will be the graph of the created triples.
   _findInIndex: function (index0, key0, key1, key2, name0, name1, name2, graph) {
-    var results = [], entityKeys = this._entitiesInverse, tmp, index1, index2;
+    var vars = (key0 ? 1 : 0) + (key1 ? 1 : 0) + (key2 ? 1 : 0);
+    var results = [], entityKeys = (vars > 1 && !graph) ? Object.keys(this._entities) : this._entitiesInverse,
+        tmp, index1, index2;
 
     // If a key is specified, use only that part of index 0.
     if (key0) (tmp = index0, index0 = {})[key0] = tmp[key0];
--- a/perf/N3Store-perf.js
+++ b/perf/N3Store-perf.js
@@ -24,7 +24,28 @@ console.timeEnd(TEST);
 
 console.log('* Memory usage for triples: ' + Math.round(process.memoryUsage().rss / 1024 / 1024) + 'MB');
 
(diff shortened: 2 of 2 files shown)


rubensworks:
As discussed last week, changing `_findInIndex` to always use the fastest variant might be the best solution here.
`Object.keys(this._entities)` for 2 or more variables and `this._entitiesInverse` for all other cases.

To be investigated...

drewp:
This patch helps my case a lot.

I have ~800 entities in my store so far:
```
      t1 = Date.now()
      for i in [0..1000]
        @graph.floatValue(@uri, U(':originTime')) // calls findByIRI(s, p), where s and p are non-null
      log('took', Date.now() - t1)

timeline.js:390 took 14654 (14ms per call!)
timeline.js:390 took 14314
timeline.js:390 took 14183
timeline.js:390 took 14003
timeline.js:390 took 14284
```

With the patch:
```
timeline.js:390 took 35 (0.035ms per call)
timeline.js:390 took 34
timeline.js:390 took 35
timeline.js:390 took 42
timeline.js:390 took 36
```

rubensworks:
I'll have some free time tomorrow, so I can a look at this then. It should be possible to make sure that lookups are optimal for all access patterns.

rubensworks:
This new version makes sure that the inverse index is only called when there are less than 2 variables,
in all other cases the original `Object.keys(this._entities)` is called.

The results below show that the new code (this PR) is faster than the original code in all cases.
(Tested with `dim = 64`)

```
OLD:
    N3Store performance test
    - Adding 262144 triples in the default graph: 323.389ms
    * Memory usage for triples: 46MB
    - Finding all 262144 triples to the default graph 4096 times (0 variables): 1755.370ms
    - Finding all 262144 triples to the default graph 8192 times (1 variable): 322.146ms
    - Finding all 262144 triples to the default graph 12288 times (2 variables): 242.523ms

    - Adding 65536 quads: 105.001ms
    * Memory usage for quads: 67MB
    - Finding all 65536 quads 16384 times: 99.361ms
NEW:
    N3Store performance test
    - Adding 262144 triples in the default graph: 614.931ms
    * Memory usage for triples: 57MB
    - Finding all 262144 triples to the default graph 4096 times (0 variables): 1538.850ms
    - Finding all 262144 triples to the default graph 8192 times (1 variable): 100.025ms
    - Finding all 262144 triples to the default graph 12288 times (2 variables): 59.296ms

    - Adding 65536 quads: 134.271ms
    * Memory usage for quads: 76MB
    - Finding all 65536 quads 16384 times: 28.999ms
```

The inverse index does however cause an increase in memory usage and triple insertion execution times.
If this would be a problem, a config option could be added to completely disable the inverse index and fall back to the original version,
this would be good for use cases where most of the lookups are done with two or more variables.

drewp:
Something may have been contaminating my test before, making the 'old' version run extra long, but the patch is still a win for that one line:

```
      t1 = Date.now()
      for l in [0..10000]
        originTime = @graph.floatValue(@uri, U(':originTime'))
      log('took', Date.now() - t1)  
```

(this is in https://github.com/drewp/light9/blob/master/light9/web/timeline.coffee#L233)

N3.js
timeline.js:361 took 4700
timeline.js:361 took 4697
timeline.js:361 took 4713
timeline.js:361 took 4671
timeline.js:361 took 4729
timeline.js:361 took 4823
...

with git pull origin pull/61/head
```
timeline.js:361 took 144
timeline.js:361 took 203
timeline.js:361 took 205
timeline.js:361 took 202
timeline.js:361 took 167
...
```

My project happens to have no less-selective queries yet, so I may be unaffected by the new code. I also have not run with data larger than a few hundred quads yet. There's going to be a lot of point-on-curve statements with very repetitive subjects and predicates.

RubenVerborgh:
I don't see the insertion performance hit as a problem, given that lookups will typically be much more common than insertions.

RubenVerborgh:
Wait wait, we're kidding ourselves here.

lib/N3Store.js
```diff
@@ -88,7 +89,9 @@ N3Store.prototype = {
   // (for instance: _subject_, _predicate_, and _object_).
   // Finally, `graph` will be the graph of the created triples.
   _findInIndex: function (index0, key0, key1, key2, name0, name1, name2, graph) {
-    var results = [], entityKeys = Object.keys(this._entities), tmp, index1, index2;
+    var vars = (key0 ? 1 : 0) + (key1 ? 1 : 0) + (key2 ? 1 : 0);
+    var results = [], entityKeys = (vars > 1 && !graph) ? Object.keys(this._entities) : this._entitiesInverse,
```
RubenVerborgh: `graph` is always truthy, so `Object.keys(this._entities)` never triggers.
(And that's for the better, because `this._entities['><'] = 0;` was never set, so it wouldn't work properly anyway.)

RubenVerborgh:
Does that mean we simply don't need the `Object.keys` solution anymore?

RubenVerborgh:
(Reverted merge for now, but can't reopen pull request unfortunately.)

rubensworks:
Ah yes, the `!graph` part was added last-minute, I didn't realize at that time that the graph could never be falsy there. That is why the test didn't fail for the missing `this._entities['><'] = 0;`.

It's still strange then that the performance test showed a speedup in all cases...
I'll look into it and open a new PR.

RubenVerborgh:
> I didn't realize at that time that the graph could never be falsy there
The graph is always a string/IRI (even the default graph has one).

BTW You can check whether all code paths are reached through `npm run coverage`.

PR #69, opened 2016-09-04
fixed querying for blank nodes in N3Store

florianthalmann:
now updating reverse index when adding blank nodes in N3Store. queries with one variable previously returned `undefined` for blank nodes. added some simple tests


RubenVerborgh:
Thanks, merged as 081d7730db7b1ea9369ea3b654375add0683853e.

PR #72, opened 2016-09-07
adding a rudementary typescript definition file

phreed:
I have been working on a set of [typescript](http://www.typescriptlang.org/) [definitions](http://www.typescriptlang.org/docs/handbook/writing-declaration-files.html) for N3.
I am planning on including them in [DefinitelyTyped](http://definitelytyped.org/) but I would prefer to include them in [this package](https://github.com/RubenVerborgh/N3.js/issues/new).

### DIFF
--- a/lib/n3.d.ts
+++ b/lib/n3.d.ts
@@ -1,0 +1,42 @@
+
+
+declare namespace N3 {
+
+    type ErrorCallback = (err: Error, result: any) => void;
+
+    interface Dictionary {
+
+    }
+
+    interface Triple {
+        subject: string,
+        predicate: string,
+        object: string
+    }
+
+    interface Output { 
+        addTriple(subject: string, predicate: string, object: string): void;
+        addTriple(triple: Triple): void;
+        end(err: ErrorCallback, result?: any): void;
+    }
+
+    interface Options {
+        format?: string,
+        prefixes?: any
+    }
+    function Writer(options: Options): Output;
+    function Writer(fd: any, options: Options): Output;
+
+    interface StoreOutput extends Output {
+        find(subject: string, predicate: string, object: string): Triple;
+    }
+    function Store(): StoreOutput;
+
+    namespace Util {
+        function createLiteral(value: any): string;
+    }
+}
+
+declare module "n3" {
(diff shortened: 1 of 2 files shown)


RubenVerborgh:
Hi @phreed, could you add some context? I'm not sure about the utility/necessity, as I'm not familiar with TypeScript. At the moment, I'd prefer this to be somewhere externally, as I cannot manage the updates of these files (not being a TypeScript user), and there seems to be no automated way to test them.

phreed:
It is better to have the typings included with the project proper rather
than in something like DefinitelyTyped

== The typings offer a partial formal specification for the API ==

Even if you do not make use of TypeScript the definitions can provide
additional hints as to how the API is intended to be used.

Consider something as simple as the 'Util.createLiteral()' function.
Does it take one argument or two?
Does the type of the first argument matter?
What are the valid values of the second argument? is it just a string?

Most of these questions are answered in your documentation but they can be
partially answered in a formal way in the TypeScript definitions.
Therefore typings are a form of documentation.

For people writing libraries a dynamic language is great but for
consumers a tightly specified API is very helpful.

In other words the typings are better thought of as part of the testsuite
rather than something to be tested by it.

RubenVerborgh:
> It is better to have the typings included with the project proper rather
than in something like DefinitelyTyped

Perhaps, but only if I can keep them up to date (I can't). The API will still undergo major changes before we reach 1.0. It's also worth noting that there is work in progress for a [shared RDF/JS interface](https://github.com/rdfjs/representation-task-force/blob/master/interface-spec.md); perhaps it's better to annotate that one.

I see and understand the utility of typing for some developers, but I have no idea how large the share of TypeScript developers is among N3.js users, and not being a TypeScript user myself, I would be adding code to my codebase which I do not understand/use/update. I think it's better not having it (or externally having it) than having it badly.

phreed:
ok
I will make the external project in DefinitelyTyped.
Thanks for the heads-up about the shared RDF/JS interface.

RubenVerborgh:
Alright. If you have a link to the entry, I'll definitely add it to the README. Thanks for your efforts!

PR #83, opened 2016-12-05
Fix browser build on windows.

DieterDePaepe:
The spawn commands used in the browser script were not working for windows, due to the use of the absolute paths it called the linux scripts instead of the `.cmd` windows variants.


RubenVerborgh:
Thanks!

PR #85, opened 2016-12-12
Move lookup of IDs to be outside graph loop in findByIRI

jimsmart:
Results in slightly cleaner code, and is more optimal in the
case of many graphs.


RubenVerborgh:
Indeed, that's how it was prior to 027c072fc51704c42770c129d81226ba7ba1c381. Thanks!

PR #87, opened 2016-12-29
N3Store API extension

jimsmart:
Additional finders:-

findGraphs(subject, predicate, object)
findGraphsByIRI(subject, predicate, object)
findSubjects(predicate, object, graph)
findSubjectsByIRI(predicate, object, graph)
findPredicates(subject, object, graph)
findPredicatesByIRI(subject, object, graph)
findObjects(subject, predicate, graph)
findObjectByIRI(subject, predicate, graph)

Iterators:-

forEach(fn, subject, predicate, object, graph)
forEachByIRI(fn, subject, predicate, object, graph)
some(fn, subject, predicate, object, graph)
someByIRI(fn, subject, predicate, object, graph)
every(fn, subject, predicate, object, graph)
everyByIRI(fn, subject, predicate, object, graph)
forGraphs(fn, subject, predicate, object)
forGraphsByIRI(fn, subject, predicate, object)
forSubjects(fn, predicate, object, graph)
forSubjectsByIRI(fn, predicate, object, graph)
forPredicates(fn, subject, object, graph)
forPredicatesByIRI(fn, subject, object, graph)
forObjects(fn, subject, predicate, graph)
forObjectByIRI(fn, subject, predicate, graph)

### DIFF
--- a/lib/N3Store.js
+++ b/lib/N3Store.js
@@ -133,7 +133,8 @@ N3Store.prototype = {
   _someInIndex: function (index0, key0, key1, key2, name0, name1, name2, graph, fn) {
     var tmp, index1, index2, varCount = !key0 + !key1 + !key2,
         // depending on the number of variables, keys or reverse index are faster
-        entityKeys = varCount > 1 ? Object.keys(this._ids) : this._entities;
+        entityKeys = varCount > 1 ? Object.keys(this._ids) : this._entities,
+        result = { subject: '', predicate: '', object: '', graph: graph };
 
     // If a key is specified, use only that part of index 0.
     if (key0) (tmp = index0, index0 = {})[key0] = tmp[key0];
@@ -151,7 +152,6 @@ N3Store.prototype = {
             var values = key2 ? (key2 in index2 ? [key2] : []) : Object.keys(index2);
             // Create triples for all items found in index 2.
             for (var l = values.length - 1; l >= 0; l--) {
-              var result = { subject: '', predicate: '', object: '', graph: graph };
               result[name0] = entity0;
               result[name1] = entity1;
               result[name2] = entityKeys[values[l]];
--- a/lib/N3Store.js
+++ b/lib/N3Store.js
@@ -139,21 +139,19 @@ N3Store.prototype = {
     // If a key is specified, use only that part of index 0.
     if (key0) (tmp = index0, index0 = {})[key0] = tmp[key0];
     for (var value0 in index0) {
-      var entity0 = entityKeys[value0];
+      result[name0] = entityKeys[value0];
 
       if (index1 = index0[value0]) {
         // If a key is specified, use only that part of index 1.
         if (key1) (tmp = index1, index1 = {})[key1] = tmp[key1];
         for (var value1 in index1) {
-          var entity1 = entityKeys[value1];
+          result[name1] = entityKeys[value1];
 
           if (index2 = index1[value1]) {
             // If a key is specified, use only that part of index 2, if it exists.
             var values = key2 ? (key2 in index2 ? [key2] : []) : Object.keys(index2);
             // Create triples for all items found in index 2.
             for (var l = values.length - 1; l >= 0; l--) {
-              result[name0] = entity0;
-              result[name1] = entity1;
               result[name2] = entityKeys[values[l]];
               if (fn(result.subject, result.predicate, result.object, result.graph)) {
                 return true;
--- a/lib/N3Store.js
+++ b/lib/N3Store.js
@@ -674,7 +674,7 @@ N3Store.prototype = {
   },
 
   forPredicatesByIRI: function (fn, subject, object, graph) {
-    var seen = {}, ids = this._ids, graphs = this._graphs, graphContents,
+    var seen = {}, ids = this._ids, graphs = {}, graphContents,
         entityKeys = this._entities;
 
     // Either loop over all graphs, or over just one selected graph.
@@ -751,7 +751,7 @@ N3Store.prototype = {
   },
 
   forObjectsByIRI: function (fn, subject, predicate, graph) {
-    var seen = {}, ids = this._ids, graphs = this._graphs, graphContents,
+    var seen = {}, ids = this._ids, graphs = {}, graphContents,
         entityKeys = this._entities;
 
     // Either loop over all graphs, or over just one selected graph.
--- a/lib/N3Store.js
+++ b/lib/N3Store.js
@@ -130,6 +130,79 @@ N3Store.prototype = {
     return results;
   },
 
+  _someInIndex: function (index0, key0, key1, key2, name0, name1, name2, graph, fn) {
+    var tmp, index1, index2, varCount = !key0 + !key1 + !key2,
+        // depending on the number of variables, keys or reverse index are faster
+        entityKeys = varCount > 1 ? Object.keys(this._ids) : this._entities;
+
+    // If a key is specified, use only that part of index 0.
+    if (key0) (tmp = index0, index0 = {})[key0] = tmp[key0];
+    for (var value0 in index0) {
+      var entity0 = entityKeys[value0];
+
+      if (index1 = index0[value0]) {
+        // If a key is specified, use only that part of index 1.
+        if (key1) (tmp = index1, index1 = {})[key1] = tmp[key1];
+        for (var value1 in index1) {
+          var entity1 = entityKeys[value1];
+
+          if (index2 = index1[value1]) {
+            // If a key is specified, use only that part of index 2, if it exists.
+            var values = key2 ? (key2 in index2 ? [key2] : []) : Object.keys(index2);
+            // Create triples for all items found in index 2.
+            for (var l = values.length - 1; l >= 0; l--) {
+              var result = { subject: '', predicate: '', object: '', graph: graph };
+              result[name0] = entity0;
+              result[name1] = entity1;
+              result[name2] = entityKeys[values[l]];
+              if (fn(result.subject, result.predicate, result.object, result.graph)) {
+                return true;
+              }
+            }
+          }
+        }
+      }
+    }
+    return false;
+  },
+
+  _lookupIndex0lookupIndex1loopIndex2: function (index0, key0, key1, fn) {
+    var index1, index2, key2;
+    // Lookup.
+    if (!(index1 = index0[key0])) return;
+    // Lookup.
+    if (!(index2 = index1[key1])) return;
+    // Loop.
+    for (key2 in index2)
+      fn(key2);
+  },
+
+  _lookupIndex0loopIndex1: function (index0, key0, fn) {
+    var index1, key1;
+    // Lookup.
+    if (!(index1 = index0[key0])) return;
+    // Loop.
+    for (key1 in index1)
+      fn(key1);
+  },
+
+  _loopIndex0lookupIndex1: function (index0, key1, fn) {
+    var key0, index1;
+    // Loop.
+    for (key0 in index0) {
+      index1 = index0[key0];
+      // Lookup.
+      if (index1[key1])
+        fn(key0);
+    }
+  },
+
+  _loopIndex0: function (index0, fn) {
+    // Loop.
+    for (var key0 in index0)
+      fn(key0);
+  },
+
   // ### `_countInIndex` counts matching triples in a three-layered index.
   // The index base is `index0` and the keys at each level are `key0`, `key1`, and `key2`.
   // Any of these keys can be `null`, which is interpreted as a wildcard.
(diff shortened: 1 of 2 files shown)


lib/N3Store.js
```diff
@@ -362,6 +427,382 @@ N3Store.prototype = {
...
+
+  findObjectsByIRI: function (subject, predicate, graph) {
+    var out = [];
+    this.forObjectsByIRI(function (o) {
+      out.push(o);
+    }, subject, predicate, graph);
+    return out;
+  },
+
+  forObject: function (fn, subject, predicate, graph) {
```
RubenVerborgh: Should probably be `forObjects`.

RubenVerborgh:
First of all: very interesting addition to the interface, thanks! I've been thinking about iteration for quite some time, but got round to implementing it. The code also looks good at a first glance, more or less like how I'd have implemented it, so that's good.

A couple of points that need to be looked at (by you, me, or others):

- [ ] **Unit tests until we hit 100% coverage.** We'll need edge cases too, but I think that, by writing them in a smart way, we can reuse code for tests just like we reuse code for implementations.
- [ ] **Check the overlap/connection with [RDF/JS](https://github.com/rdfjs/representation-task-force/blob/master/interface-spec.md).**
- [ ] **Check if we can shorten/optimize code in some cases.** For instance, the necessity of `expandPrefixedName` strikes me. Maybe it's silly to have two variants for every function. We could have perhaps generators for the other functions. Maybe the `IRI` variants should be the default.
- [ ] **Consider removing the existing array interface (or build it on top of `forEach`).**
- [ ] **Check performance, especially if following the previous suggestion.**

dmitrizagidulin:
@jimsmart what's the difference between findGraphs() and findGraphsByIRI()?

jimsmart:
@dmitrizagidulin all of the ByIRI methods simply skip the expansion from short names (e.g. rdf:type) to full IRI - I just followed the pattern of the existing API here.

@RubenVerborgh Thank you very much for the prompt code review! :) 

Unit tests / 100% coverage, totally agree - I have tests that give good coverage on my Go quad store, which has a surprisingly similar API ;) it's somewhat time consuming writing tests, but it's not so bad once you get into it. I think I can probably find some time next week to port those tests over to Javascript - if nobody else gets there before me?! (I'm usually ok writing tests - but the multiple permutations on an API of this kind do get tedious)

I've briefly looked at RDF/JS in the past, but not spent any time with it. Somebody else is perhaps better suited for such a comparison. Perhaps I should just ask what is the goal here? Some kind of compatibility maybe? Or are we just doing a comparison to see where the API stands, feature-wise?

It's your call regarding the ByIRI methods — my initial thoughts are that it's useful to have both, depending on what kind of code one is writing. I guess part of the decision lies with whether one is trying to mimic / implement the W3C specs (which, IMHO, both the 'RDF Interfaces' spec and the higher-level 'RDF API' both leave a lot to be desired - but, by their own admission, they are both unfinished specs, so hey ho). Keeping both versions of the methods seems like a good solution: it's not like keeping the extras around is a problem - beyond making the API fatter by a factor of two - let's face it: they are not complex code. I don't think I understand what you mean by generators here? (Javascript is not my primary language, so forgive me if I'm missing something obvious).

Yes, one can easily piggy-back the existing find() to use forEach - although: I tried this, and found that Javascript's callbacks and/or the V8 JIT wouldn't make it perform as well as the existing code, which is why I left that as it is. 

I've found that both callbacks and 'tricky code' seem to hurt performance (in general for Javascript, as far as I can tell). But my only benchmark is running the scripts in the perf folder, under Node/V8. Some aspects of Javascript performance tuning seems like voodoo to me, despite having done plenty of performance optimisation in the past in other languages - mostly due to the craziness of JIT I suspect.

Slight aside: is there a better way of measuring performance in Javascript? (Recently I've been spoiled because Go has pretty good benchmarking in the standard library - but I'm quick to admit that I don't really know the lay of the land in Javascript world, yet)

Speaking of performance, I have a couple of random related comments/observations which I'll throw into the pot...

I originally had a different implementation of _someByIndex, which followed my Go code more closely. Curiously, on my machine, it ran the 'find 1m quads with 0 params' a whole 10 seconds faster (~50s down from ~60s, IIRC), but it slowed down the find with 1 and 2 params benchmark. I never got to the bottom of it, and re-jigged my code. The only take-away we really have from that is that there is definitely more performance to be eeked out, should we choose to follow that route. My suspicion is that to gain performance in all areas (i.e. with 0/1/2 params) it's probably easiest to have specific methods for _findInIndex for the different key combinations (doing ifs for whether-we-have-a-param-or-wildcard really ought not to be in that loop for max performance - those same checks are done a layer above before we call-in anyhow!) - if we took this route, we'd end up with functions similar (but different, obvs) to the ones in my patch e.g. _findGivenKeyYandKeyZ. There's a maximum of eight combinations, but because we have 3 indexes, I suspect we'd end up with less than eight methods (though I may be wrong on this latter point).

My other observation regarding performance is that when I profiled a run-through of the existing perf test (I'm only talking about the N3Store test here), I saw that we spend 2.5% of the time doing regex. I *suspect* that we may get better performance doing that ourselves (it can only be the short-name to IRI expansion here? Correct me if I'm wrong) - but I didn't look into that area particularly, and again, my suspicions may be wrong (I'm good at being wrong sometimes, doh!)

Cheers.

lib/N3Store.js
jimsmart: Yes of course - good catch

jimsmart:
Whilst I remember:

The behaviour of every() in pretty much every Javascript API I've looked at (I'm really talking about every() on basic things like Array, etc) is somewhat sucky, IMO. I'm sure it's all down to legacy code and backwards compatibility. But my issue is basically this:-

A call to every() will return true when called on an empty collection / empty iteration set. The docs always go to great pains to explain this... MDN say in their docs for Array.every: "every acts like the "for all" quantifier in mathematics. In particular, for an empty array, it returns true. (It is vacuously true that all elements of the empty set satisfy any given condition.)" — which is all well and good, having an explanation for what is basically counter-intuitive behaviour. But in the real-world, this is as much use as a chocolate teapot. 

Sure, if one is relying on the return result from every() when working with an array, it's not difficult to do the extra test (is the array empty). But it's trickier when working with a filtered selection of quads - well, not really tricky, but one has to do a somewhat-expensive count() afterwards.

It's my belief that this counter-intuitive behaviour actually comes from the fact that implementations of every() just use the code for some() with extra negations - which is exactly what the code in this patch set currently does.

But in my Go quad store I recently decided to ditch all that higher-maths explanation for somewhat cranky behaviour, and I added-in the simple check to give a less surprising / more useful result from the function - the end result being: in my every(), if the given callback was never called, then it will returns false.

The code difference is negligible - it's the three lines in this short method that mention the flag 'some'
https://github.com/jimsmart/store4/blob/master/quadstore.go#L426

(Obviously my Go implementation is a different kettle of fish: our standard library is a different beast compared to anything in Javascript. We don't have every() so I can do as I please in my API providing I document it)

What do others think of this whole issue? Should we keep 'every' as it is (it returns true for an empty iteration set), just like all other Javascript everys do? — Or should we go with what I believe is more useful and less surprising? (Perhaps this behaviour is not surprising to Javascript veterans, because it's what they are used to?)

— Comments/thoughts on this matter are most appreciated :)

lib/N3Store.js
jimsmart: Fixed locally - will be in my next push.

jimsmart:
Okey dokey... 

- I've added tests to get 100% coverage
- Found and fixed a couple of bugs in my code along the way.
- Modified every() to be the 'more useful' implementation, mentioned above.
- Found a minor bug in size(), added a test, then fixed it.
- Merged your recent changes.
- Brought my code in line with your changes re. handling of default graph, and use of xxId naming in methods.

Todo:

- Documentation of new methods (any volunteers?)
- Possible code sharing between forEach and find

jimsmart:
Previously I said

> There's a maximum of eight combinations, but because we have 3 indexes, I suspect we'd end up with less than eight methods (though I may be wrong on this latter point).

— I've taken this route in my Go QuadStore: turns out it's only four short methods. 

It gained me just over 10% performance improvement. 

Benchmark details in the commit message, if anyone's interested https://github.com/jimsmart/store4/commit/dea111ceb51d7ccde6a64a86ac9855955b516989#diff-b65b82ca9ece0afb303b9ddef3630ee2

Cheers

RubenVerborgh:
This is excellent, thank you so much. Can't believe you went through the trouble of making all these tests. Really great stuff.

Please give me some time to inspect and test, but I'll surely merge!

RubenVerborgh:
Hah, and I just found a neat trick to implement `_findInIndex` and `_someInIndex` as a single function, removing the code duplication of the existing array variants.

jimsmart:
> This is excellent, thank you so much.

You're welcome. Thank you for a great library!

jimsmart:
Note that there are many API calls I do not explicitly test, my tests are somewhat lazy and rely on inside-knowledge, e.g. I test `findSubjects`, knowing that it calls `findSubjectsByIRI`, which then calls `forSubjectsByIRI`, so I skip those. Similarly `forEach` gets tests for all permutations of input, whereas `some` and `every` just get enough to ensure their return values are ok (because `forEach` proxies to `some`) — So if you go doing any really crazy refactoring at any point, try to remember these points!

jimsmart:
Hey, can you tell me about this:

```
  var ..., varCount = !key0 + !key1 + !key2,
        // depending on the number of variables, keys or reverse index are faster
        entityKeys = varCount > 1 ? Object.keys(this._ids) : this._entities;
```
— I mean: I can see what it's doing, and I've proven the statement true by benchmarking, but what kind of voodoo is this exactly? O_o

After much thought, my only guess is that it's _time-taken-to-convert-object-to-array-followed-by-n-lookups_ vs _time-taken-by-n-map-lookups_.

Much Googling didn't shed any light on the matter.

Having only run the perf tests under Node, I wonder if this still holds true for JS engines other than V8 ...?

RubenVerborgh:
> Note that there are many API calls I do not explicitly test: my tests are somewhat lazy and rely on inside-knowledge, e.g. I test findSubjects, knowing that it calls findSubjectsByIRI, which then calls forSubjectsByIRI, so I skip those.

No worries, as long as there's 100% coverage, that's fine. A coverage drop would tell me if something becomes not covered after refactoring.

> After much thought, my only guess is that it's time-taken-to-convert-object-to-array-followed-by-n-lookups vs time-taken-by-n-map-lookups.

Yeah, that was also our conclusion. This line came after an investigation why half of the cases were faster than the others, and it turned out that the hybrid solution made both fast. Even if we're unsure about the exact cause, fast is fast :smile:

> Having only run the perf tests under Node, I wonder if this still holds true for JS engines other than V8

Likely not, however, as long as they don't cause terrible performance on the other engines, that's alright with me. Optimization for V8 is our primary objective, as we target Node.js in the first place (but compatibility with others is important too).

RubenVerborgh:
I'm reviewing and testing in [feature-store-api-extension](https://github.com/RubenVerborgh/N3.js/tree/feature-store-api-extension). See ed002c9f1f956ac72f261ff0b2a8fb1c3acdc8dd for the `_findInIndex` and `_someInIndex` trick. I'll try to apply it in a couple of other places as well.

jimsmart:
Nice little trick! Yeah, much better to do that, than a callback, when collecting results.

jimsmart:
Regarding the `*ByIRI` methods, another option which we've not discussed would be to separate all of those calls out into another layer entirely. 

So you'd have a vanilla `N3Store` that takes full IRIs, and a wrapper/delegate/proxy version, with an identical API, which performed all the IRI expansion in every call.

I guess there's pros and cons either way.

RubenVerborgh:
Yes, that layer is exactly what I was thinking: as soon as prefixes are
set, it becomes active. Users can then choose between convenience or speed.

aaronhelton:
I've been making use of this branch today for my own purposes, and these additions are really useful. One thing that's missing, however, unless I'm missing it elsewhere, is a means of grabbing the preferred label(s) of a subject, the way Python's RDFLib graph module does, including an optional language code for easier filtering. 

If I'm overlooking something, let me know.

RubenVerborgh:
Just to let you know I'm still on this. I'm just going through a crazy busy period right now, but this will make it into N3.js master for the next major release.

RubenVerborgh:
Okay, after huge delays from my side, the code should be ready in 01b01509f124a620143ba57febff9c96a00e6d2a.

Now I just need to document the interface, and can release as 0.10.

RubenVerborgh:
Merged as e2ed8e1e1a4e473b699891705bc094af84259e0d, released as v0.10.0.

Thanks so much for this contribution and your patience.

jimsmart:
Thank you!

I like what you did with _uniqueEntities — and I think renaming find* to get* was also a good move.

PR #89, opened 2017-01-07
N3 store perf experiment

jimsmart:
This is the result of an experiment. It's not really a pull request, although you're welcome to take the code if you wish :)

So... I wrote the 'four functions' version of `some`, and hooked up `find` to use `some`.

My benchmarks for current `master` are typically:

```
N3Store performance test
- Adding 16777216 triples to the default graph: 18528.181ms
* Memory usage for triples: 709MB
- Finding all 16777216 triples in the default graph 65536 times (0 variables): 88729.043ms
- Finding all 16777216 triples in the default graph 131072 times (1 variable): 5624.881ms
- Finding all 16777216 triples in the default graph 196608 times (2 variables): 5504.780ms

- Adding 16777216 quads: 29252.079ms
* Memory usage for quads: 673MB
- Finding all 16777216 quads 1048576 times: 25739.851ms
```

And the benchmarks for this branch can be in the region of:

```
$ node N3Store-perf.js
N3Store performance test
- Adding 16777216 triples to the default graph: 21008.823ms
* Memory usage for triples: 719MB
- Finding all 16777216 triples in the default graph 65536 times (0 variables): 39437.696ms
- Finding all 16777216 triples in the default graph 131072 times (1 variable): 4961.082ms
- Finding all 16777216 triples in the default graph 196608 times (2 variables): 7438.038ms

- Adding 16777216 quads: 27909.561ms
* Memory usage for quads: 799MB
- Finding all 16777216 quads 1048576 times: 16721.452ms
```

Hmmm. Curious indeed.

The problem function in this new experimental code is `_indexSomeGivenKey0` (called for '2 variables', above). Regarding the voodoo trick: my benchmarks seem to show that it doesn't matter whether I use the `Object.keys` or `this._entities` here.

- I get very inconsistent benchmark times.
- Partly I think this may be due to my laptop turbo-boosting or throttling. Don't know. It's not the best machine to benchmark on really because of this.
- Partly I think it may be because the benchmark is actually timing a large chunk of work compared to my previous optimisation experience?
- Partly I think it may be because it's an interpreted language. But I am biased, being much more familiar with compiled languages.
- It's so ridiculously fast on the other methods/permutations that I'm beginning to think we've definitely hit a weird edge case in V8 (take a look at `_indexSomeGivenKey0`... it couldn't do any less - except for not having a callback if we were doing `find`, I guess. But seriously?) Maybe the Chrome team would be interested?
- Apologies in advance for all the commented-out dead code
- Some of the code is plain ugly - and no apologies could really cover that. 

Anyhow, make of it what you will. (Good luck)

— Have you perhaps considered implementing the index lookups in asm.js? <gets coat and runs>

### DIFF
--- a/lib/N3Store.js
+++ b/lib/N3Store.js
@@ -187,6 +187,149 @@ N3Store.prototype = {
       fn(key0);
   },
 
+  // Lookup, lookup, lookup
+  _indexSomeGivenAllKeys: function (index0, key0, key1, key2, name0, name1, name2, graph, fn) {
+    var varCount = !key0 + !key1 + !key2,
+        // depending on the number of variables, keys or reverse index are faster
+        entityKeys = varCount > 1 ? Object.keys(this._ids) : this._entities;
+
+    // It turns out that varCount is actually a count of null parameters.
+    // We always have a null parameter count of 0.
+    var entityKeys = this._entities;
+
+    // console.log('given all keys, varCount ', varCount);
+
+    // Lookup
+    if (key0 in index0) {
+      var index1 = index0[key0];
+      // Lookup
+      if (key1 in index1) {
+        var index2 = index1[key1];
+        // Lookup
+        if (key2 in index2) {
+          var entity0 = entityKeys[key0];
+          var entity1 = entityKeys[key1];
+          var entity2 = entityKeys[key2];
+          var result = { subject: '', predicate: '', object: '', graph: graph };
+          result[name0] = entity0;
+          result[name1] = entity1;
+          result[name2] = entity2;
+          // We can just return the fn result - we're not looping
+          return fn(result);
+        }
+      }
+    }
+    return false;
+  },
+
+  // Lookup, lookup, loop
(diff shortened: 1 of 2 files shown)


jimsmart:
Travis hates my lack of adherence to linting - but eslint hated my code :(

jimsmart:
Having given this a little thought now: if anything, perhaps it's worth keeping just `_indexSomeWhenGivenAllKeys` from this branch - that seems to be a very big win for small additional code / code changes. 

Performance vs code duplication / readability / maintenance, often a tricky call.

jimsmart:
One problem with benchmarks is that they can make you focus on the wrong things...

Looking through my code here, I don't have even a single call to `some` / `find` / etc with _all_ of the parameters given. 

I don't believe it's a common use case - it's every other case we want optimising - I don't think that case is worthwhile optimising for. 

— Even if it does look good on the benchmark.

RubenVerborgh:
Indeed—to be looked at when we have better real-world benchmarking scenarios.

PR #90, opened 2017-01-10
Added 'use strict' directive

jimsmart:
This patch set adds `"use strict";` to the top of every file (except perf and tests).

I believe strict mode has numerous advantages, including making some aspects of development more pleasant.

All tests still pass. 

Nothing else needed changing (which is not always the case when adding strict mode to existing codebases!)

### DIFF
--- a/N3.js
+++ b/N3.js
@@ -1,3 +1,4 @@
+"use strict";
 // Replace local require by a lazy loader
 var globalRequire = require;
 require = function () {};
--- a/lib/N3Lexer.js
+++ b/lib/N3Lexer.js
@@ -1,3 +1,4 @@
+"use strict";
 // **N3Lexer** tokenizes N3 documents.
 var fromCharCode = String.fromCharCode;
 var immediately = typeof setImmediate === 'function' ? setImmediate :
--- a/lib/N3Parser.js
+++ b/lib/N3Parser.js
@@ -1,3 +1,4 @@
+"use strict";
 // **N3Parser** parses N3 documents.
 var N3Lexer = require('./N3Lexer');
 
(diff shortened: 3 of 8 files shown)


RubenVerborgh:
> I believe strict mode has numerous advantages, including making some aspects of development more pleasant.

Could you elaborate a bit on this? I haven't found the need for this myself, but I'd be happy to reconsider.

jimsmart:
I was thinking particularly about how strict mode turns some errors that are normally silent (and therefore hard to chase down) into hard-errors. 

I'm pretty sure I lost an hour (or more) to exactly this when I was implementing the store API extension? Can't fix that without a time-machine though! ;)

I suspect part of the reason you've not seen the need is that you may be more familiar with JavaScript than I am - Another reason is that you have your eslint settings quite strict, already catching a bunch of stuff that strict mode disallows.

But I notice this change caused failed builds on Travis and AppVeyor?

RubenVerborgh:
> I was thinking particularly about how strict mode turns some errors that are normally silent (and therefore hard to chase down) into hard-errors.

Do you have a couple of examples? This could convince me!

> I suspect part of the reason you've not seen the need is that you may be more familiar with JavaScript than I am

Well, everything that helps better code, I'm willing to consider!

> But I notice this change caused failed builds on Travis and AppVeyor?

Yeah, the linter currently demands that strict mode is off; I can change that.

jimsmart:
It's not something I feel strongly about: if you feel it's not necessary, no big deal :) 

I believe the current lint settings catch all the same issues anyhow (variables escaping to global, duplicate props, trying to set props on a primitive).

Maybe review the situation again at a later date: it's a 2-min edit should you change your mind later :)

RubenVerborgh:
Okay, will review at later date 😄

PR #91, opened 2017-01-14
isIRI returns true for an object

jmatsushita:


RubenVerborgh:
Thanks for preparing this!

However, the N3.js code indeed assumes that the `isIRI` etc. methods are called with the correct type, i.e., strings. It only really makes sense to call `isIRI` with possible candidates for `true`/`false`, not with objects where we know it is going to be `false` (such as objects). Moreover, the introduced check is incomplete, as `irIRI(new String('http://example.org/'))` should give `true` but does not anymore. While this could be fixed, this would add more type checking code, potentially introducing a performance penalty.

So summarizing: thanks, but JavaScript being a duck-typed language, I don't think that `isIRI` and the likes should have explicit type checking.

jmatsushita:
@RubenVerborgh Thanks for explaining. It makes complete sense :)

PR #106, opened 2017-05-05
added simple toString serialization feature and associated tests

royludo:
It would be really helpful to have a quick, simple and synchronous way to get the store's content to string. So I added 2 exposed functions in the Writer, tripleToString and triplesToString (do change the names, I was lacking inspiration there).
tripleToString is almost the same as _writeTripleLine, it's copy pasted except it directly returns a string. triplesToString calls tripleToString on an array.

I added 5 necessary tests and corrected another one.

There's a possibility of refactoring _writeTripleLine by making it call tripleToString, but I didn't touch it.

Lint and tests passed.

### DIFF
--- a/lib/N3Writer.js
+++ b/lib/N3Writer.js
@@ -98,10 +98,7 @@ N3Writer.prototype = {
     delete this._prefixMatch;
     // Write the triple
     try {
-      this._write(this._encodeIriOrBlankNode(subject) + ' ' +
-                  this._encodeIriOrBlankNode(predicate) + ' ' +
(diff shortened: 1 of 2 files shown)


RubenVerborgh:
Cool!

> There's a possibility of refactoring _writeTripleLine by making it call tripleToString, but I didn't touch it.

Makes sense, could you please do that as well?

royludo:
@RubenVerborgh done !

RubenVerborgh:
Thanks! Merged as f29db1b444ab527b4a9d84c2b64769e9321938ae.

Issue #109, opened 2017-06-13
Improve information in error object


LaurensRietveld:
Did you consider adding more information or context to an error object?
E.g. add the token information (as json), as property to the error. This way we don't have to parse the error msg string and extract this information ourselves.

A bonus (though this is probably more suited as another issue) is to add the context (e.g. the 100 chars before, and 100 chars after) to a particular error

RubenVerborgh:
> Did you consider adding more information or context to an error object?

I should!

> E.g. add the token information (as json), as property to the error. This way we don't have to parse the error msg string and extract this information ourselves.

+1, simple fix, big gain

> A bonus (though this is probably more suited as another issue) is to add the context (e.g. the 100 chars before, and 100 chars after) to a particular error

After certainly not a problem; before more tricky. Related to #102.

LaurensRietveld:
I'll submit a PR that adds the line and (if available) as props to the error object. 
Just 2 questions:
- I'll just naively add the props to the err object (i.e., I won't be subclassing and stuff, seems overkill)
- Did you think about using prettier instead of the linter? (finding the linter annoying tbh ;)). Downside will be that most of the line will be marked as modified, i.e. you'll lose history for blaming and stuff

RubenVerborgh:
> I'll submit a PR that adds the line and (if available) token as props to the error object.

Great, added minor comments.

> I'll just naively add the props to the err object (i.e., I won't be subclassing and stuff, seems overkill)

Yes please :-)

> Did you think about using prettier instead of the linter? (finding the linter annoying tbh ;)).

What you call the “annoying” linter is actually [my highly personalized linter configuration](https://github.com/RubenVerborgh/N3.js/blob/master/.eslintrc), which matches the style I want to write 😄 

But I'm open to feedback if there are particular annoyances.

LaurensRietveld:
hehe, nothing in particular about the style that's annoying. Just like the tooling to fix all the errors for me instead

RubenVerborgh:
Following up in #110.

PR #110, opened 2017-06-14
Add context information to thrown errors

LaurensRietveld:
The PR following our discussion on #109

### DIFF
--- a/lib/N3Lexer.js
+++ b/lib/N3Lexer.js
@@ -360,7 +360,9 @@ N3Lexer.prototype = {
   // ### `_syntaxError` creates a syntax error for the given issue
   _syntaxError: function (issue) {
     this._input = null;
-    return new Error('Unexpected "' + issue + '" on line ' + this._line + '.');
+    var err = new Error('Unexpected "' + issue + '" on line ' + this._line + '.');
+    err.line = this._line;
+    return err;
   },
 
 
--- a/lib/N3Lexer.js
+++ b/lib/N3Lexer.js
@@ -319,9 +319,10 @@ N3Lexer.prototype = {
       }
 
       // Emit the parsed token
-      callback(null, { line: line, type: type, value: value, prefix: prefix });
+      var token = { line: line, type: type, value: value, prefix: prefix };
+      callback(null, token);
       this._prevTokenType = type;
-
+      this._prevToken = token;
       // Advance to next part to tokenize
       input = input.substr(matchLength || match[0].length, input.length);
     }
@@ -361,7 +362,11 @@ N3Lexer.prototype = {
   _syntaxError: function (issue) {
     this._input = null;
     var err = new Error('Unexpected "' + issue + '" on line ' + this._line + '.');
-    err.line = this._line;
+    err.context = {
+      token: undefined,
+      line: this._line,
+      previousToken: this._prevToken,
+    };
     return err;
   },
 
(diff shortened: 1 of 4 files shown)


LaurensRietveld:
I added some context checks to the unit tests. Doing it everywhere seemed a bit overkill, so I added it to a select few.

lib/N3Parser.js
```diff
@@ -775,8 +775,11 @@ N3Parser.prototype = {
   // ### `_error` emits an error message through the callback
   _error: function (message, token) {
     var err = new Error(message + ' on line ' + token.line + '.');
-    err.token = token;
-    err.line = token.line;
+    err.context = {
+      token: token,
+      line: token.line,
+      previousToken: this._lexer._prevToken,
```
LaurensRietveld: I'm not sure whether this line might become an issue. The token is passed as argument, where the previous token is fetched from the lexer directly. Is there a scenario where both are not 'in sync'?

LaurensRietveld: ps. afaik this doesnt seem to be an issue in the unit tests I updated

RubenVerborgh:
Thanks, will take this up soon!

lib/N3Lexer.js
```diff
@@ -319,9 +319,10 @@ N3Lexer.prototype = {
       }
 
       // Emit the parsed token
-      callback(null, { line: line, type: type, value: value, prefix: prefix });
+      var token = { line: line, type: type, value: value, prefix: prefix };
+      callback(null, token);
       this._prevTokenType = type;
-
+      this._prevToken = token;
```
RubenVerborgh: Should check the performance impact (should be minimal).

lib/N3Parser.js
```diff
@@ -774,7 +774,13 @@ N3Parser.prototype = {
 
   // ### `_error` emits an error message through the callback
   _error: function (message, token) {
-    this._callback(new Error(message + ' on line ' + token.line + '.'));
+    var err = new Error(message + ' on line ' + token.line + '.');
+    err.context = {
+      token: token,
+      line: token.line,
+      previousToken: this._lexer._prevToken,
```
RubenVerborgh: Alternatively, the parser might store it. Would avoid accessing a private variable. (Also, a little strange that the lexer exposes `_prevToken` but not `_token`.)

RubenVerborgh:
Added a couple of minor things to think about.

lib/N3Lexer.js
LaurensRietveld: did some eyeballing on the N3Lexer-perf script (ran on an ntriple file with 4million statements). Memory footprint is the same. User time _might_ have decreased, but with around 1 or 2% at most

lib/N3Parser.js
LaurensRietveld: You mean by adding the prevToken as a parameter in the tokenize callback? (https://github.com/OpenTriply/N3.js/blob/8ccffc54904a3f5a5b669f8f6f434c22b2ac1746/lib/N3Parser.js#L897)
I'm comfortable with just accessing the private variable tbh

LaurensRietveld:
Added some responses. Using it now btw in some of my libs. Works great!

RubenVerborgh:
Thanks, merged as 4601c2a8d5e54a1d9f4d53b3ca33f4f87dfe9c16.

PR #113, opened 2017-08-07
change conflicting variable name in Writer

ericprud:
This simple [HTML interface using rawgit](https://jsfiddle.net/ericP/2anpxbrx/1/) gets an error (hit [Parse] to see it) with any literal with `[\r\n\t\b]` in it because the variable `escapeReplacements` appears in [N3Lexer](https://github.com/RubenVerborgh/N3.js/blob/master/lib/N3Lexer.js#L9) and [N3Writer](https://github.com/RubenVerborgh/N3.js/blob/master/lib/N3Writer.js#L13). Renaming it to `turtleEscapes` in `N3Writer` eliminates this error as seen in this [HTML interface using this PR](https://jsfiddle.net/ericP/r77ye4wL/).


RubenVerborgh:
Thanks, merged as e67f6793f2c9bbb98e99644d526a8278827f0dc2. Renamed the variable to `escapedCharacters`, since it's not just for Turtle.

PR #114, opened 2017-08-07
N3 incorrectly allows multi-line single-quoted literals

LaurensRietveld:
Added a unit test that fails on this particular issue


RubenVerborgh:
Wow, weird that this only shows up now. I need to check this, thanks.

PR #118, opened 2017-10-13
Feature RDF/JS

rubensworks:
To do:
* [x] RDFJS data: N3Writer & N3StreamWriter
* [x] RDFJS data: N3Store
* [x] RDFJS data: N3Util


rubensworks:
`N3Writer` has been changed in a backwards-compatible way, the only cost of this is a simple `typeof term === 'string'` check on each term, so performance impact should be minimal.

I can still change it if you prefer a complete breaking change, in which case more invasive internal changes to the writer will need to be done.

RubenVerborgh:
Thanks @rubensworks, breaking changes are fine with me. v1.0.0 will not be backward compatible.

rubensworks:
@RubenVerborgh I removed backwards compatibility with the string-based interface of N3Writer. However, internally everything currently still works with the old string-based representation. I believe this will have better performance than using the RDF/JS datastructures for everything internally as well.

rubensworks:
@RubenVerborgh Everything should be compatible with the RDF/JS interface now.

I had to make some design decisions, such as converting everything to a string-based representation internally, and the way `N3Util.expandPrefixedName` works.
Please let me know if you agree with these.

elf-pavlik:
Have you tried running the test suite mentioned in https://github.com/rdfjs/representation-task-force/issues/47#issuecomment-294256920 ?

rubensworks:
@elf-pavlik Thanks for the suggestion, I'll look into it soon.

rubensworks:
It looks like the [test suite](https://github.com/bergos/rdf-data-model-test-suite) is built around `DataFactory`. This PR however does not provide a `DataFactory` implementation, so this can't be tested, unless it should?
@RubenVerborgh Thoughts?

bergos:
@rubensworks the test suite moved to [rdf-data-model](https://github.com/rdf-ext/rdf-data-model/). Old documentation is still valid.

pietercolpaert:
Is performance worse or similar with the new interface? @rubensworks

RubenVerborgh:
Slightly worse last time I tested, but acceptable.

pietercolpaert:
Would this PR make N3.js ready for a new major release?

RubenVerborgh:
Yes, this will become v1.0.0 when finished. I just desperately need 1 full
day to work on this.

l00mi:
gogogo!

RubenVerborgh:
Thanks for your help, @rubensworks. I've heavily modified the interfaces, so the TypeScript definitions will be obsolete. I've removed usage of the internal representation from most places, and many helper methods from Util are gone.

rubensworks:
Great! I will look into updating the TS typings soon.

PR #130, opened 2018-03-22
N3Writer: Support literals with spec-conformant languages

gremid:
While processing an LOD export of the [Getty's Art & Architecture Thesaurus (AAT)](http://www.getty.edu/research/tools/vocabularies/aat/), I ran into an issue with literals accompanied by language codes that appear to be spec-compliant but are not supported by `N3Writer`.

Please review this fix and consider merging. Thanks!

### DIFF
--- a/lib/N3Writer.js
+++ b/lib/N3Writer.js
@@ -1,7 +1,7 @@
 // **N3Writer** writes N3 documents.
 
 // Matches a literal as represented in memory by the N3 library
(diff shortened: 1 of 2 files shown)


RubenVerborgh:
Good catch, the original code was too restrictive compared to [what the parser accepts](https://github.com/RubenVerborgh/N3.js/blob/v0.11.2/lib/N3Lexer.js#L55).

gremid:
Yes, I adapted the regex from the lexer in order to make the writer work.

RubenVerborgh:
We've got a weird unrelated build failure though. Let me check.

PR #131, opened 2018-04-13
replaced .id with RDF/JS .value in prefix handling

bergos:


RubenVerborgh:
Right!

RubenVerborgh:
Amended this slightly into f9d4b3cdf462c884eb8bfd3b404cfeb11378a43a, such that we're not creating the named node twice anymore.

Issue #149, opened 2018-10-30
Add baseIRI option to documentation of N3.Parser


pheyvaer:
It is possible to add the baseIRI to the constructor of the N3.Parser, but this is not documented in the README. It would be nice to have this added.

PR #150, opened 2018-11-02
Add documentation about base IRI of N3.Parser

pheyvaer:
I added an example in the documentation that shows how to add the base IRI to the N3.Parser (#149)


RubenVerborgh:
Thanks, merged as 9fa26a2ec12b9992531d64d0fc62b495c3579a89!

PR #151, opened 2018-11-09
Added test where chunk is split in middle of non-breaking space

rubenw:
When a non-breaking space is split across different chunks in a stream, it is not parsed correctly by the StreamParser. The result contains "��" instead of the non-breaking space " ". I added a (skipped) test for this.


RubenVerborgh:
Wow. Very good find and test.

RubenVerborgh:
Thanks a lot. Merged as 458ac6a474c4c8281ae535b786c4b8d45934cbf5 and fixed in 7a955d76f6abfa30917425762320382caba79abb.

PR #160, opened 2019-01-24
RDFJS Source interface on N3Store

iddan:
[Source](http://rdf.js.org/#source-interface) interface for Store implementation
Based on https://github.com/rdfjs/N3.js/pull/161


iddan:
Ready for review

RubenVerborgh:
Rebased on latest master, will review now.

lib/N3Stream.js
```diff
@@ -0,0 +1,5 @@
+var Stream = require('stream').Stream;
+
+class N3Stream extends Stream {}
+
+module.exports = N3Stream;
```
RubenVerborgh: This class is trivial; should not be included.

test/N3Store-test.js
```diff
@@ -1036,6 +1037,38 @@ describe('N3Store', function () {
...
+      var stream = store.match();
+      var data = [];
+      stream.on('data', (quad) => {
+        data.push(quad);
+      });
+      store.addQuad('null', 'null', 'null', 'null');
+      data.should.deep.equal([new Quad('null', 'null', 'null', 'null')]);
+    });
+
+    it('should emit quad every time a matching quad is added to the store', function () {
```
RubenVerborgh: That's not how I understand the `Source` interface. Is it stated somewhere that the stream should keep on updating indefinitely?

lib/N3Store.js
```diff
@@ -19,6 +21,8 @@ class N3Store {
     this._entities = Object.create(null); // inverse of `_ids`
     // `_blankNodeIndex` is the index of the last automatically named blank node
     this._blankNodeIndex = 0;
+    // '_changeStream' holds active streams created by this.match()
+    this._changeStream = new N3Stream();
```
RubenVerborgh: Not sure about the necessity; see below.

test/N3Store-test.js
iddan: I guess you can interpret it like that. In RDFLib you can listen to additions and removals of the store. If that's not the intended API I can change it.

RubenVerborgh:
I don’t think its the intended API; listening to changes would be a different mechanism.

iddan:
I will propose a different API.

PR #161, opened 2019-01-24
Refactor classes to ES6 Class syntax

udielenberg:


RubenVerborgh:
Nice, thanks!

We should probably do this for all classes in the project, for consistency. Is this in your pieline, or should we find other volunteers?

udielenberg:
I'll do it ✌️

RubenVerborgh:
Thanks, I appreciate it!

iddan:
All classes have been refactored

iddan:
It can be merged even though @udielenberg hasn't updated the title yet

iddan:
Breaking Changes:
 - Constructors can't be called without a `new` keyword as specified by ES6
 - DefaultGraph is a singleton object and not a class. This reduces the magic required for keeping one instance of the class.

iddan:
I tried, but none of the ways I've tried worked.

RubenVerborgh:
This does the trick:

```JavaScript
let instance;

class Singleton {
  constructor() {
    return instance || (instance = this);
  }
}

console.log(new Singleton() === new Singleton());
```

iddan:
Cool, let me see if I can make it work

iddan:
Seems like it breaks many of the tests. For some reason when taking this approach `this` becomes undefined. I think it's a correct move to make DefaultGraph an object and not a constructor which returns the same object. I'm unaware of any standard library that acts that way

RubenVerborgh:
Thanks for trying, will look at it.

RubenVerborgh:
Thanks a lot, and sorry for the delay.

PR #162, opened 2019-01-25
RDFJS Source interface remove event

iddan:
Adds `remove` event to Source's Stream
Based on https://github.com/rdfjs/N3.js/pull/160


RubenVerborgh:
Rebased on current #161.

RubenVerborgh:
Implemented in https://github.com/rdfjs/N3.js/pull/174; thanks for your help.

PR #164, opened 2019-02-20
A suggestion to a bit more info on bundling

megoth:
As suggested by @RubenVerborgh. Uncertain if the wording is good enough though.


RubenVerborgh:
Thanks a lot!

PR #168, opened 2019-03-20
RDFJS Source interface on N3Store

iddan:
Source interface for Store implementation
Based on #161
Reopen of #160


RubenVerborgh:
Implemented in https://github.com/rdfjs/N3.js/pull/174; thanks for your help.

PR #170, opened 2019-03-27
tell folks to create a UMD bundle

ericprud:
Since there's no variable assigned, just module-exports, folks need to supply a global variable name.

(took me a few to figure this out so i hope to save others the time.)


RubenVerborgh:
Thanks!

PR #171, opened 2019-03-29
+ instructions, scripts and testbed for browser products

ericprud:
I wanted to shorten the learniing curve for using N3.js in a browser.

Remaining issues:
1. I screwed around reading webpack docs for while then gave up to create a hack `N3-webpack.js` to emualte the behavior of `browserify -s N3`
2. The browser/simple.html page doesn't demonstrate reading from `ReadableStreams` as I never figured out how to. Other than that, it covers pretty much everything in the README.md.

### DIFF
--- a/N3-webpack.js
+++ b/N3-webpack.js
@@ -1,0 +1,1 @@
+N3 = require('./N3.js')
--- a/README.md
+++ b/README.md
@@ -38,8 +38,29 @@ const N3 = require('n3');
 
 N3.js seamlessly works in browsers via [webpack](https://webpack.js.org/)
 or [browserify](http://browserify.org/).
-If you're unfamiliar with these tools,
-you can read
+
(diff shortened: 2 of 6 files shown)


RubenVerborgh:
Would it be okay if we pointed at this issue from the README, or if we put this in the wiki?

I'm asking because the code would need some editing to fit with the rest of the repo.

ericprud:
I think I was supposed to have done something here. Since I didn't, what should we do now?

RubenVerborgh:
We could wiki it, or I'd need to spend some time integrating it with the repo.

I don't mind doing the latter, it's just that my open source issue queue has become long lately, so it will take me a bit of time to get to it.

RubenVerborgh:
Pausing this for now; feel free to revive at any point.

Issue #145, opened 2018-08-24
RDFJS Source interface on N3Store


rubensworks:
It would be nice if N3Store would implement the [Source](http://rdf.js.org/#source-interface) interface.
This would make it possible to for example use N3Store as a direct back-end in comunica, instead of [having to wrap it](https://github.com/comunica/comunica/tree/master/packages/actor-init-sparql-rdfjs#usage).

This interface defines the match method as follows:
```
Stream match(optional (Term or RegExp) subject, optional (Term or RegExp) predicate, optional (Term or RegExp) object, optional (Term or RegExp) graph);
```
Unfortunately, this includes regexes, so this can't be implemented efficiently in N3Store (unless for terms that are _not_ a regex)

rubensworks:
Yep, that should be an easy an clean way to do it!

It might be possible to optimize a bit and apply the regexes directly on each index level, to avoid iterating over (potentially large) index values for terms that don't match anyways. But that won't be as straightforward to implement, so this could be future work™.

iddan:
Started working on it at https://github.com/rdfjs/N3.js/pull/160

RubenVerborgh:
Implemented in https://github.com/rdfjs/N3.js/pull/174.

PR #174, opened 2019-04-16
Implement RDFJS Store (and Source) interface on N3Store

rubensworks:
As requested in #145.

Note that Source#match used to accept RegExp as terms, but this has been removed now from the RDFJS spec, which simplifies implementation a lot.

I saw that #162 and #168 also offer a (partial) implementation of this, but they seem to have stalled a bit, and I need support for this ASAP, so I created this new PR (which is a bit less invasive).


RubenVerborgh:
Thanks, @rubensworks, great work.

RubenVerborgh:
v1.1.0

PR #175, opened 2019-04-23
add support for periods in local IRI part

bryonjacob:
Turtle syntax allows for periods (`'.'`) in the local part of a prefixed IRI, as long as it's not in the leading or terminal position:

https://dvcs.w3.org/hg/rdf/raw-file/default/rdf-turtle/index.html#grammar-production-PN_LOCAL

This PR modifies the regex that matches IRIs which can be shortened to allow for periods as appropriate, and updates the `should use prefixes when possible` test to account for this.

_note: there are still more things which are permissible by the spec that N3.js will not shorten, this only addresses the inclusion of periods in permissible locations in an IRI local part, not the entire spec_

### DIFF
--- a/lib/N3Writer.js
+++ b/lib/N3Writer.js
@@ -229,7 +229,7 @@ class N3Writer {
       }
       IRIlist = IRIlist.replace(/[\]\/\(\)\*\+\?\.\\\$]/g, '\\$&');
       this._prefixRegex = new RegExp('^(?:' + prefixList + ')[^\/]*$|' +
-                                     '^(' + IRIlist + ')([a-zA-Z][\\-_a-zA-Z0-9]*)$');
+                                     '^(' + IRIlist + ')([a-zA-Z](?:[\\-_.a-zA-Z0-9]*[\\-_a-zA-Z0-9])?)$');
     }
     // End a prefix block with a newline
     this._write(hasPrefixes ? '\n' : '', done);
--- a/test/N3Writer-test.js
+++ b/test/N3Writer-test.js
@@ -135,11 +135,15 @@ describe('N3Writer', function () {
     it('should use prefixes when possible',
       shouldSerialize({ prefixes: { a: 'http://a.org/', b: 'http://a.org/b#', c: 'http://a.org/b' } },
                       ['http://a.org/bc', 'http://a.org/b#ef', 'http://a.org/bhi'],
+                      ['http://a.org/bc.d', 'http://a.org/b#ef', 'http://a.org/bhi'],
+                      ['http://a.org/bc.', 'http://a.org/b#ef', 'http://a.org/bhi'],
                       ['http://a.org/bc/de', 'http://a.org/b#e#f', 'http://a.org/b#x/t'],
                       ['http://a.org/3a', 'http://a.org/b#3a', 'http://a.org/b#a3'],
                       '@prefix a: <http://a.org/>.\n' +
                       '@prefix b: <http://a.org/b#>.\n\n' +
                       'a:bc b:ef a:bhi.\n' +
+                      'a:bc.d b:ef a:bhi.\n' +
+                      '<http://a.org/bc.> b:ef a:bhi.\n' +
                       '<http://a.org/bc/de> <http://a.org/b#e#f> <http://a.org/b#x/t>.\n' +
                       '<http://a.org/3a> <http://a.org/b#3a> b:a3.\n'));
 


RubenVerborgh:
Thanks. However, my worry is about performance, since this regex looks more expensive. Can we check that performance is not degraded?

bryonjacob:
Are there any existing performance-related tests or benchmarks for N3Writer that could be extended to compare against for this?  

I suspect that the  "or" block at the beginning of the regex that includes all the referenced prefixes is the long pole here and that the added production is minimal in cost...

rubensworks:
Untested, but this regex may be a bit cheaper:

`'^(' + IRIlist + ')([a-zA-Z]([\\-_.a-zA-Z0-9]*[\\-_a-zA-Z0-9])?)$')`

RubenVerborgh:
> Are there any existing performance-related tests or benchmarks for N3Writer that could be extended to compare against for this?

Unfortunately not; I only see benchmarks for parser and store.
However, for this case, I guess a simple JSperf over some URLs would do.

But yes, in general, the reason why not all URI are abbreviated is purely performance. It's feasible to make an exhaustive regex or check, but it would be too slow.

bryonjacob:
understood - I get the performance concerns - 

I will look into JSperf, and test it against the original, my suggested regex and your suggested improvement, and submit whichever one is more performant (or recant the PR if the performance degrades unacceptably...)

thanks!

bryonjacob:
https://jsperf.com/compare-n3-js-regexes

Initial testing shows that my regex is about 7% slower (from 970K ops/sec to 900K, on my laptop+browser) and yours was about 10% slower (which surprised me, analytically I'd have expected yours to be similar or better, too)

After some tweaking, I added one more variant:
`'^(' + IRIlist + ')([a-zA-Z][\\-_.a-zA-Z0-9]*[\\-_a-zA-Z0-9]|[a-zA-Z])$'`

That one is maybe a little better on average - still somewhere in the 5-7% slowdown depending on the run

bryonjacob:
Thinking about this case a bit more - there seem to be a couple ways to look at it:
- this performance degradation is acceptable, and worth the tradeoff to handle shortening more names (helps me with my current issue, obviously - but it's somewhat arbitrary to support '.' and not the rest of the spec...)
- performance concerns trump this level of enhancement, and the N3Writer class should not adopt this at the measured expense

other than potentially "accepting this PR as-is", there's a couple of options:
- update this PR to be able to shorten ALL URLs with valid local parts (per Turtle spec), and measure the performance impact of doing so.  Maybe it's not worth a 7% degradation on this regex performance to just add periods, but it'd be worth a 10% degradation to support the whole spec?  Also, this degradation is measured ONLY against the regex match - the whole operation consists of more moving parts, and if we measured the performance impact measured against the whole cost of serialization might be much smaller.
- add a configuration option to N3Writer to specify the more aggressive shortening algorithm, leaving the default behavior as-is - so that one can choose to shorten as many URLs as possible at a performance cost.

I'd be willing to pursue either of those options, but before putting any more work into this wanted to check with the owners of this project to get a sense of which of those would be preferred, and might be likely to be accepted if completed?  

A third option is to create a separate TurtleWriter class and release it as a separate component, but this seems like functionality that would make sense for other N3.js users, so I wanted to contribute it back if possible.  Thanks!

RubenVerborgh:
It's a hard one indeed. Especially because we've found (in other testing) that streaming and decoding compressed N-Triples is faster than Turtle.

The performance differences are however considerable (10% on my machine) for a case that might not be all too common (correct me if I'm wrong). I remember that dots were not even allowed in earlier Turtle drafts. Given that valid, spec-compatible output is produced in both cases, I'm inclined to not do it.

Your suggestion for a flag does make sense. I'm just thinking about use cases. Are you pretty printing for human consumption? A wise man once told me one could do a PhD on Turtle pretty printing, which is likely true 🙂

bryonjacob:
no doubt, Turtle pretty-printing is a complex beast.

I'm not sure how common dots are in local URI parts, to be honest - one internal spec for how we coin IRIs uses them liberally, and the TTL files that contain these URIs are routinely read by humans, so it matters to ME, but I can't speak for the general population.

I agree that compressed n-triples is the best route to take for general purpose state transition between systems - I use Turtle only when I have a use case where humans will need to consume the output (which almost always necessitates that the files are relatively small, so the performance impact of this regex, which still runs in around a microsecond, are not concerning)

Given all of this - if you think it's a reasonable option, I will spend some time to re-work this into an "option flag" that will cause serialized URIs to be "shortened to the greatest degree the spec allows" (or as close as we can reasonably get - I haven't tried writing the operation yet :) - we can document the performance tradeoff one could expect when activating that flag, and leave the end user the choice.  If you don't think there's enough value in adding the flag, I can easily write and maintain a separate writer for my purposes, I just wanted to contribute this back to N3.js if it made sense to do so...  

Thanks!

RubenVerborgh:
Thanks for suggesting! Such a flag sounds great!

Let me close this PR then, and then we can consider one with a flag.

PR #179, opened 2019-06-10
+ N3Store.extractLists()

ericprud:
`N3Parser` emits first/rest lists. `N3Writer.list(members)` serializes rdf:Collections. `sequesterLists()` removes the first/rest lists and returns a Map() than can be used to call `N3Writer.list()`.

Issues:
1. could use `{}` instead of `Map()` but I have the impression that es is trying to migrate to Map()s.
2. could be called `sequesterCollections()` or `isolateCollections()` to emphasize that it works with `rdf:Collections`. OTOH, that's a lot of extra characters.


RubenVerborgh:
Thanks, @ericprud.

To be very precise, I presume it only returns the longest possible lists? So given `(1 2 3)` it will just return `(1 2 3)` and not additionally `()`, `(3)`, `(2 3)`? (Relates a bit to https://github.com/rdfjs/N3.js/pull/180#issuecomment-502486972).

ericprud:
Right, it actually can't take a shorter one because it makes sure that there's exactly one ref to the list head ([co-references out of head](https://github.com/rdfjs/N3.js/pull/179/files#diff-4653bcb1247373577f3037a492ec16fdR1221) and [co-references into head](https://github.com/rdfjs/N3.js/pull/179/files#diff-4653bcb1247373577f3037a492ec16fdR1236)) and that there are no extra rdf:{first,rest} (all the rest of the tests).

ericprud:
Justification for the name: `sequesterLists()` actually removes the triples rather than just hiding them (e.g. https://github.com/rdfjs/N3.js/pull/179/files#diff-4653bcb1247373577f3037a492ec16fdR1133 had 5 triples before sequesterList() was called). OTOH, they're still implied in the Collection data structure returned by sequesterList() so maybe they're just hidden.

I'm not married to the function name, but the functionality is pretty handy.

RubenVerborgh:
Sorry for the massive delay, @ericprud.

I have refactored the code into `#extractLists({ remove: false, ignoreError: false }) throws Error`.
This makes the interface more consistent with the rest of the code, but loses the error collection.

@ericprud @jairo-campos-JD @james-martin-jd can you let me know whether this suits your needs?

james-martin-jd:
Looks good to me, good call on the name change.

PR #180, opened 2019-06-10
Writer list heads

ericprud:
Follows up on [#179] by taking the result of `extractLists()` as an N3Writer constructor options:
``` javascript
const lists = graph.extractLists({ remove: true });
const writer = new N3.Writer({ lists });
writer.addQuads(graph.getQuads());
```
Otherwise, the user has to do something like:
``` javascript
const listHeads = sequesterLists(graph)
const writer = new N3.Writer()
writer.addQuads(graph.getQuads().map(q => {
  let members = listHeads.get(q.subject.value)
  if (members)
    q.subject = writer.list(members)
  members = listHeads.get(q.object.value)
  if (members)
    q.object = writer.list(members)
  return q
}))
```

### DIFF
--- a/src/N3Writer.js
+++ b/src/N3Writer.js
@@ -33,6 +33,7 @@ export default class N3Writer {
     if (outputStream && typeof outputStream.write !== 'function')
       options = outputStream, outputStream = null;
     options = options || {};
+    this._lists = options.lists;
 
     // If no output stream given, send the output as string through the end callback
     if (!outputStream) {
@@ -129,8 +130,12 @@ export default class N3Writer {
   // ### `_encodeIriOrBlank` represents an IRI or blank node
   _encodeIriOrBlank(entity) {
     // A blank node or list is represented as-is
-    if (entity.termType !== 'NamedNode')
+    if (entity.termType !== 'NamedNode') {
+      // If it is a list head, pretty-print it
+      if (this._lists && (entity.value in this._lists))
+        entity = this.list(this._lists[entity.value]);
       return 'id' in entity ? entity.id : '_:' + entity.value;
+    }
     // Escape special characters
     var iri = entity.value;
     if (escape.test(iri))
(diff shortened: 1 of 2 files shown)


RubenVerborgh:
Thanks for the work, @ericprud.

This one scares me a bit, as the writer is streaming. What guarantee do we have that intermediary nodes of the list are not being used? Or is passing in `listHeads` such a guarantee?

ericprud:
Such a list would be malformed. RDF [does *not* define well-formedness](https://www.w3.org/TR/rdf11-mt/#rdf-collections) but [OWL does](https://www.w3.org/TR/2012/REC-owl2-mapping-to-rdf-20121211/#Mapping_from_RDF_Graphs_to_the_Structural_Specification).

Procedurally, callers of this function would have to behave well or they would trick the writer into losing co-references into malformed lists. The accompanying #179 enforces well-formedness across a Store. It is intended to be used with the Writer like so:
``` javascript
const writer = new N3.Writer({
  prefixes: { "": IRI_this, ui: NS_Ui, dc: NS_Dc },
  listHeads: quadStore.sequesterLists() # removes well-formed lists from the store.
})
writer.addQuads(quadStore.getQuads())
```

RubenVerborgh:
Depends on #179.

RubenVerborgh:
Example code:

```JavaScript
const { Parser, Store, Writer } = require('n3');

const turtle = `
<a> <b> (<c> <d> <e>).
<f> <g> (<h> <i> <j>).
`;

const store = new Store();
new Parser().parse(turtle, (err, quad) => {
  if (quad)
    store.addQuad(quad);
  else {
    const lists = store.extractLists({ remove: true });
    const writer = new Writer({ lists });
    writer.addQuads(store.getQuads());
    writer.end((err, output) => console.log(output));
  }
});
```

PR #184, opened 2019-09-21
README: adds a an explanation for the automatic prefixing of blank node labels and the blankNodePrefix constructor argument

jacoscaz:


RubenVerborgh:
Excellent, thanks!

PR #188, opened 2019-12-02
allow empty relative IRIs

bergos:
#182 added support for relative IRIs, but it doesn't cover the case where the IRI is empty. An error was thrown, that the IRI was invalid. This PR fixes the problem by:

- use `null` and not an empty string for invalid IRIs
- adding a test case for empty relative IRIs

From my point of view using `null` for invalid IRIs makes it more clear that something is wrong. So this PR could be also seen as a very small refactoring with the side effect of supporting empty relative IRIs :-)

### DIFF
--- a/src/N3Parser.js
+++ b/src/N3Parser.js
@@ -29,7 +29,7 @@ export default class N3Parser {
     this._supportsQuads = !(isTurtle || isTriG || isNTriples || isN3);
     // Disable relative IRIs in N-Triples or N-Quads mode
     if (isLineMode)
-      this._resolveRelativeIRI = function (iri) { return ''; };
+      this._resolveRelativeIRI = function (iri) { return null; };
     this._blankNodePrefix = typeof options.blankNodePrefix !== 'string' ? '' :
                               options.blankNodePrefix.replace(/^(?!_:)/, '_:');
     this._lexer = options.lexer || new N3Lexer({ lineMode: isLineMode, n3: isN3 });
@@ -154,7 +154,7 @@ export default class N3Parser {
     case 'IRI':
     case 'typeIRI':
       var iri = this._resolveIRI(token.value);
-      if (iri === '')
+      if (iri === null)
         return this._error('Invalid IRI', token);
       value = this._namedNode(iri);
       break;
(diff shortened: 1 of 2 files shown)


RubenVerborgh:
Thanks!

Are relative IRIs compatible with https://rdf.js.org/data-model-spec/#namednode-interface ? It doesn't mention anything about them being absolute or relative, but a lot of code I've seen seems to assume absolute.

bergos:
It was discussed and we agreed to keep it open especially for such use cases. [Here](https://github.com/rdfjs/data-model-spec/issues/57) is the related issue in the spec repo.

RubenVerborgh:
> [Here](https://github.com/rdfjs/data-model-spec/issues/57) is the related issue in the spec repo.

The issue itself seems to be more about valid versus invalid IRIs.

> It was discussed and we agreed to keep it open especially for such use cases. 

Okay, but I am still worried about the assumption of absolute URIs being made by existing consumers.

Would it make sense to have it as a flag on the parser (that defaults to the current behavior)? Like:

```
    if (isLineMode || requireBaseIri)
      this._resolveRelativeIRI = function (iri) { return null; };
```

bergos:
> The issue itself seems to be more about valid versus invalid IRIs.

The [second comment ](https://github.com/rdfjs/data-model-spec/issues/57#issuecomment-200349782) about the LDP use case is exactly what I want to handle with this PR.

> Okay, but I am still worried about the assumption of absolute URIs being made by existing consumers.

Even without this PR the parser would create already relative IRIs. This PR only fixes the case were the relative IRI is an empty string.

I hope this was not added by accident. In the previous PR I added [this test](https://github.com/rdfjs/N3.js/blob/master/test/N3Parser-test.js#L897-L907), but with the wrong label. I think it was initially something like "should not try to resolve the IRIs and just use the relative IRI", but I forgot to remove the "not" after changing it.

> Would it make sense to have it as a flag on the parser (that defaults to the current behavior)?

The `requireBaseIri` flag would make the parsing more strict and default is the most relaxed parsing mode? I think that would be consistent with the other flags. But maybe a different name. `requireBaseIri` sounds like it would also throw an error if there is no base IRI (content or argument), even if all IRIs in the content are absolute. Maybe something like `disableRelativeIRIs`?

RubenVerborgh:
> Even without this PR the parser would create already relative IRIs.

Right, apologies for not checking #182, which I had forgotten about.

> Maybe something like `disableRelativeIRIs`?

Actually, in hindsight, we could've had `baseIRI: null` for #182 to explicitly request this behavior. But we can't introduce that now without changing major version. So I will accept as-is.

Thanks for your work on this!

PR #193, opened 2020-02-09
Accepts literal as subject

sroze:
This makes possible to have literals as subjects; which is what the EYE reasoner is using to describe its mathematical proofs.

### DIFF
--- a/test/N3Parser-test.js
+++ b/test/N3Parser-test.js
@@ -1063,6 +1063,10 @@ describe('Parser', function () {
     it('should not parse a formula as list item',
       shouldNotParse(parser, '( <a> { <b> <c> <d> } <e> ).',
         'Unexpected graph on line 1.'));
+
+    it('should not parse a literal as subject',
+      shouldNotParse(parser, '1 <a> <b>.',
+        'Unexpected literal on line 1.'));
   });
 
   describe('A Parser instance for the TriG format', function () {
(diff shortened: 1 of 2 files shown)


RubenVerborgh:
Thanks for taking this on!
However, literals as a subject are only allowed in N3, not in other syntaxes such s as Turtle. As a result, the spec tests are currently broken (see Travis logs above).

sroze:
@RubenVerborgh thank you for the review! Seems like wrapping it all within a `if (this._n3Mode)` statement is not enough... Do you have any suggestion? 🤔

test/N3Parser-test.js
```diff
@@ -1506,6 +1506,18 @@ describe('Parser', function () {
 
     it('should not parse an invalid ^ path',
       shouldNotParse(parser, '<a>^"invalid" ', 'Expected entity but got literal on line 1.'));
+
+    it('should parse literal as subject',
+        shouldParse(parser, '<a> <evidence> ( <e> [ a <fact>; <gives> {1 <greaterThan> 0}]).',
```
RubenVerborgh: I'd suggest a more simple case here to test.

RubenVerborgh:
> Seems like wrapping it all within a `if (this._n3Mode)` statement is not enough... Do you have any suggestion?

They were temporary test failures because of GitHub Pages; fixed now.

test/N3Parser-test.js
sroze: Indeed; added a much simpler one 👍

sroze:
@RubenVerborgh thank you for the review. I've added the support for typed & language-specific literals 👍

sroze:
@RubenVerborgh it’s all ready for your eyes :)

RubenVerborgh:
Thanks, will look ASAP.

RubenVerborgh:
Thanks for the hard work, and for your patience!

PR #194, opened 2020-02-09
Allow formula as list item

sroze:
N3 allows formula as list item; this adds the support for it.


RubenVerborgh:
Thanks a lot!

RubenVerborgh:
Thanks to this PR, we now pass 1024 / 1234 of N3 tests, up from 954 previously 🎉

Issue #197, opened 2020-04-07
Issue with blank node ID counter


martinmaillard:
I was having issues with conflicting blank node IDs after parsing some quads with *@rdfjs/parser-n3* and then manually adding more quads to the dataset.

After some research, I found out that it was caused by the fact that **N3.js** maintains its own [blank node ID counter](https://github.com/rdfjs/N3.js/blob/master/src/N3Parser.js#L51) and doesn't let the data factory handle it.

```javascript
const ParserN3 = require('@rdfjs/parser-n3')
const $rdf = require('rdf-ext')

const stream = fs.createReadStream(filePath) // file contains multiple blank nodes
const parser = new ParserN3({ factory: $rdf }) // passing a custom factory
const dataset = await $rdf.dataset().import(parser.import(stream))

$rdf.blankNode().value // == 'b1'
```

Is there a specific reason to have a separate blank node ID counter in *N3.js* or would it be possible to let the data factory generate new IDs?

RubenVerborgh:
> Is there a specific reason to have a separate blank node ID counter in _N3.js_ or would it be possible to let the data factory generate new IDs?

Oh, that would be the remains of the pre-factory situation 🙂 No need indeed for the parser to create this.

martinmaillard:
Would it help if I submit a PR to fix it?

RubenVerborgh:
Feel free; although I expect some trouble with the tests. There, you probably want to inject a factory where you can control the blank node ID, so the tests themselves don't have to change.

If it doesn't work out, happy to have a look myself.

martinmaillard:
I'm giving it a try but am indeed having issues with so called "named blank nodes", which have a specific logic to keep their original name in the parsed quads.

PR #198, opened 2020-04-16
Let the data factory handle blank node IDs

martinmaillard:
This is an attempt at fixing #197

### DIFF
--- a/src/N3DataFactory.js
+++ b/src/N3DataFactory.js
@@ -6,6 +6,7 @@ const { rdf, xsd } = namespaces;
 
 var DataFactory, DEFAULTGRAPH;
 
+var _blankNodePrefix = 'n3-';
 var _blankNodeCounter = 0;
 
 // ## Term constructor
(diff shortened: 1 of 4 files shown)


src/N3DataFactory.js
```diff
@@ -268,6 +269,8 @@ DataFactory = {
     Triple: Quad,
     fromId,
     toId,
+    _resetBlankNodeIds: () => { _blankNodeCounter = 0; },
+    _setBlankNodePrefix: (prefix) => { _blankNodePrefix = prefix; },
```
martinmaillard, line 273: I'm not sure it makes sense to be able to change the blank node prefix. I added this because the tests seemed to expect it.

RubenVerborgh: Might still be useful for identification purposes later on (when merging triples from multiple sources).

RubenVerborgh:
Thanks a lot!

PR #201, opened 2020-05-01
Replace setImmediate with setTimeout

NoelDeMartin:
It seems like [setImmediate](https://developer.mozilla.org/en-US/docs/Web/API/Window/setImmediate) is not recommended for production code. I found about this when I realized it isn't available in worker threads.

One approach would be to create a helper method, but given that it isn't used a lot in the project I think the best approach is just to use setTimeout.

### DIFF
--- a/src/N3Lexer.js
+++ b/src/N3Lexer.js
@@ -437,7 +437,7 @@ export default class N3Lexer {
       this._input = input;
       // If a callback was passed, asynchronously call it
       if (typeof callback === 'function')
-        setImmediate(function () { self._tokenizeToEnd(callback, true); });
+        setTimeout(function () { self._tokenizeToEnd(callback, true); }, 0);
       // If no callback was passed, tokenize synchronously and return
       else {
         var tokens = [], error;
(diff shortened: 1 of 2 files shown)


RubenVerborgh:
> It seems like [setImmediate](https://developer.mozilla.org/en-US/docs/Web/API/Window/setImmediate) is not recommended for production code.

For browser code. But N3.js also needs to work in Node, and `setTimeout` is slow.

Since browsers don't have `setImmediate`, so you can use a shim of your choice instead (including a `setTimeout` shim).

NoelDeMartin:
I see, I wasn't aware of that.

Having into account how simple the shim is, would you consider including it with the library? I don't mind having it in my projects, but it can save others some potential confusion.

RubenVerborgh:
I have no problem including it, but I'm not currently shipping with any browser configuration. So that would be a dependency. Maybe it can be a note in the README.

Not that where it really matters (https://github.com/rdfjs/N3.js/pull/201/files#diff-4437ff81da62d687c649e35228ccd752L1017), the shim is there already; the other one is just an edge case that will rarely happen multiple times per second, let alone at the point where it would make a tangible performance difference.

NoelDeMartin:
Ok I'll close this in favor of #202 to have clean commit history.

RubenVerborgh:
For future reference: I instead moved to microtasks: ed3a787ec3bb348281230d4d3d2a334621fdc9b1

PR #207, opened 2020-07-14
Added the QuadTerm type and conversion methods

simonvbrae:


src/N3DataFactory.js
```diff
@@ -200,6 +202,12 @@ export function termFromId(id, factory) {
     return factory.literal(id.substr(1, endPos - 1),
             id[endPos + 1] === '@' ? id.substr(endPos + 2)
                                    : factory.namedNode(id.substr(endPos + 3)));
+  case '<':
+    // Parse quad
+    let result = id.match(/^<<("(?:""|[^"])*"[^ ]*|[^ ]+) ("(?:""|[^"])*"[^ ]*|[^ ]+) ("(?:""|[^"])*"[^ ]*|[^ ]+) ?("(?:""|[^"])*"[^ ]*|[^ ]+)?>>$/);
+    result = result.splice(1).filter((x) => x !== undefined).map((id) => termFromId(unEscape(id), factory));
```
RubenVerborgh, line 208: The `splice`/`filter`/`map` combination creates three array loops; better to have a `for` (starting from 1)

RubenVerborgh: The undefined check is better done as `typeof x !== 'undefined'`. (Compilers are optimized for this.)

src/N3DataFactory.js
```diff
@@ -256,6 +265,53 @@ export class Quad {
...
+
+  // ### The text value of this literal
+  get value() {
+    return this.quad;
+  }
+}
+
+// ### Creates a string from a Quad
+function quadToId(quad) {
+  let result = [quad.subject, quad.predicate, quad.object].map(termToId).map(escape).join(' ');
```
RubenVerborgh, line 288: Let's avoid the creation of three arrays and be a bit more redundant with string interpolation, just for perf reasons.

src/N3DataFactory.js
```diff
@@ -256,6 +265,53 @@ export class Quad {
...
+
+  // Omit default graph
+  if (!isDefaultGraph(quad.graph)) {
+    result += ` ${termToId(quad.graph)}`;
```
RubenVerborgh, line 292: …and would add this to the single string interpolation then (hoping that it does not become too long of a line, but we could add returns inside of `${…}`)

src/N3DataFactory.js
```diff
@@ -256,6 +265,53 @@ export class Quad {
...
+  return result;
+}
+
+// ### Escapes the quotes within the given literal
+function escape(id) {
+  return escapeHelper(id, /"/g, '""');
+}
+
+// ### Unescapes the quotes within the given literal
+function unEscape(id) {
```
RubenVerborgh, line 303:
```suggestion
function unescape(id) {
```

src/N3DataFactory.js
```diff
@@ -256,6 +265,53 @@ export class Quad {
...
+  return escapeHelper(id, /""/g, '"');
+}
+
+function escapeHelper(id, regex, replacement) {
```
RubenVerborgh, line 307: How about simply (warning, untested):

```javascript
return id.replace(/^"(.*".*)(?="[^"]*$)/, (_, quoted) => `"${quoted.replace(regex, replacement)}`);
```

Note how the regex requires the presence of a quote before substitution if even considered (and the tail replacement is a non-captured lookahead).

RubenVerborgh: I don't like the name `escapeHelper`, but if it is indeed possible to make it a one-liner as above, we could just simply put the first regex in a constant, and then inline the replacements in the `escape` and `unescape` functions.

test/N3DataFactory-test.js
```diff
@@ -143,4 +143,31 @@ describe('DataFactory', function () {
...
   });
+
+  describe('quadTerm', function () {
+    it('should return a quadTerm with the correct value', function () {
+      DataFactory.quadTerm(new Quad(
+        new NamedNode('http://ex.org/a'),
+        new NamedNode('http://ex.org/b'),
+        new Literal('abc'),
+        new DefaultGraph()
+      )).value
```
RubenVerborgh, line 154: @rubensworks We are violating object homomorphism here (`value` not a string); I'm afraid this will slow things down. We likely want to reconsider at the RDF/JS level.

rubensworks: Good point, I didn't realise this.

RubenVerborgh:
Obsoleted by #210 if I understand correctly.

PR #210, opened 2020-07-16
Added RDF* support by extending Quad from Term

simonvbrae:
This PR implements RDF* on the data level. The changes include:
- Additional tests
- Make Quad extend Term
- Added the field 'value' to Quad which has the empty string as its value
- Added the field 'termtype' to Quad which has the value 'Quad'
- Updates to the termFromId and termToId methods that allow conversion from RDF* objects to string and back.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -17,6 +17,8 @@ import {
   DefaultGraph,
   Quad,
   Triple,
+  escape,
+  unescape,
 
   termFromId,
   termToId,
@@ -41,6 +43,8 @@ export {
   DefaultGraph,
   Quad,
   Triple,
+  escape,
+  unescape,
 
   termFromId,
   termToId,
--- a/src/index.js
+++ b/src/index.js
@@ -17,8 +17,6 @@ import {
   DefaultGraph,
   Quad,
   Triple,
-  escape,
-  unescape,
 
   termFromId,
   termToId,
@@ -43,8 +41,6 @@ export {
   DefaultGraph,
   Quad,
   Triple,
-  escape,
-  unescape,
 
   termFromId,
   termToId,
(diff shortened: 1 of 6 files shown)


src/index.js
```diff
@@ -41,6 +43,8 @@ export {
   DefaultGraph,
   Quad,
   Triple,
+  escape,
+  unescape,
```
RubenVerborgh, line 47: I wouldn't expose them; they're highly internal functions for ID conversion.

simonvbrae: Agreed. I thought this would be useful for end-users but actually there are no realistic use cases I can think of, will solve.

RubenVerborgh:
@simonvbrae / @rubensworks Just checking, which RDF/JS proposal is this implementing now?

simonvbrae:
> @simonvbrae / @rubensworks Just checking, which RDF/JS proposal is this implementing now?

The one from this pull request in the data-model-spec: https://github.com/rdfjs/data-model-spec/pull/165

larsrh:
I appreciate the change, but for the record, this broke client code despite only being included in a minor version bump.

I have code that accepts a `RDF.Term | RDF.Quad`. I used to be able to distinguish them by checking presence of the `termType` property. This doesn't work anymore.

RubenVerborgh:
@larsrh Thanks for letting us know. It’s semver minor though, as we are adding functionality that does not break an existing contract. The absence of a property is not part of the RDF/JS contract and should not be relied on. For instance, N3.js terms will also expose `.id`, but its absence or presence should not be relied upon, as it’s not in the contract.

It is a good point though that we probably want a distinction check that does not involve enumerating all possible term types. Let’s take that up in https://github.com/rdfjs/data-model-spec/pull/165

larsrh:
That's true, but the TypeScript compiler will rely on it (when the `termType` prop is absent, tsc narrows the type to `RDF.Quad`). Arguably tsc's behaviour is not part of the contract, so I fully understand the trade-off.

PR #213, opened 2020-07-27
Extend writer for RDF* Quads

simonvbrae:
This PR adds support for writing nested quads to the Writer component of N3. The changes include
* Added a check in the method that encodes objects
* Added a method for converting subjects
* Changed the Writer to handler Quads
* Added tests

### DIFF
--- a/test/N3Writer-test.js
+++ b/test/N3Writer-test.js
@@ -570,15 +570,17 @@ describe('Writer', function () {
       });
     });
 
-    it('should serialize a triple with a triple as subject', function () {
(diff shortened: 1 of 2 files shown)


test/N3Writer-test.js
```diff
@@ -569,6 +569,62 @@ describe('Writer', function () {
         done();
       });
     });
+
+    it('should serialize a triple with a triple as subject', function () {
```
rubensworks, line 573: Make sure to always use different names for your tests, some test runners don't like this.

PR #214, opened 2020-07-28
Adding documentation about RDF*

simonvbrae:
This PR aims to add documentation detailing the support of RDF* to the README file. This first version might be good but I'm open for suggestions.

There are some things I'm not sure about so I'll write them down quickly:
- As a link to RDF* I used [Olaf Hartig's position statement](https://blog.liu.se/olafhartig/2019/01/10/position-statement-rdf-star-and-sparql-star/), maybe another URL is necessary.
- I didn't mention the extensions we made to RDF* such as quads within `<< >>` yet because I was not sure where to put it, and maybe it's not necessary to mention this at all.

### DIFF
--- a/README.md
+++ b/README.md
@@ -339,13 +339,13 @@ In addition, the N3.js parser also supports [Notation3 (N3)](https://www.w3.org/
 The N3.js parser and writer are also fully compatible with the RDF* variants 
 of the W3C specifications.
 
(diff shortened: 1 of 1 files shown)


rubensworks:
I would also mention how to enable RDF* parsing/writing, i.e., setting the `format`.

PR #215, opened 2020-07-29
Add 'termType' property to Quad.toJSON() result

AlexeyMz:
This allows to distinguish Quad `subject` (or `object`) value by looking at `termType` when serializing Quad to JSON.


RubenVerborgh:
Indeed, thanks!

PR #223, opened 2020-10-22
Complete Use of DataFactory with termFromId

### LINKED ISSUE #222: Missing a Use of Custom DataFactory in N3 Store
The `termFromId` function from the `DataFactory` is called with the resolved DataFactory implemention of the N3 Store in all cases except the one in `_uniqueEntities`. This makes it so that the terms in the store could be a mix of N3 DataFactory terms and the custom DataFactory terms.

This is a simple fix. I'll work on a PR for this.

dhurlburtusa:
Ensure DataFactory, potentially a custom implementation, is used throughout the store.

Fixes #222.

### DIFF
--- a/src/N3Store.js
+++ b/src/N3Store.js
@@ -198,11 +198,11 @@ export default class N3Store {
   // ### `_uniqueEntities` returns a function that accepts an entity ID
   // and passes the corresponding entity to callback if it hasn't occurred before.
   _uniqueEntities(callback) {
-    var uniqueIds = Object.create(null), entities = this._entities;
-    return function (id) {
+    var uniqueIds = Object.create(null);
+    return id => {
       if (!(id in uniqueIds)) {
         uniqueIds[id] = true;
-        callback(termFromId(entities[id]));
+        callback(termFromId(this._entities[id], this._factory));
       }
     };
   }


dhurlburtusa:
I am not sure what the issue is with the failed test. Please let me know if there is something I need to do on my end.

Note: All unit tests in all environments pass.
Note: All specs in all environments pass except one.

Seems like there is a missing build artifact that causes the failure.

RubenVerborgh:
@dhurlburtusa Thanks!

@rubensworks Seems like a failure of rdf-test-suite on Node 15?

src/N3Store.js
```diff
@@ -198,11 +198,11 @@ export default class N3Store {
   // ### `_uniqueEntities` returns a function that accepts an entity ID
   // and passes the corresponding entity to callback if it hasn't occurred before.
   _uniqueEntities(callback) {
-    var uniqueIds = Object.create(null), entities = this._entities;
+    var uniqueIds = Object.create(null), entities = this._entities, factory = this._factory;
```
RubenVerborgh, line 201: I maybe wonder if we want to convert the below to an arrow function instead? Then we don'tt need to store all the variables.

rubensworks:
@RubenVerborgh I assume it's a Node 15 bug (given that it's just out).
It's failing here: https://github.com/rdfjs/N3.js/blob/master/spec/parser.js#L1
Which should be correct.

RubenVerborgh:
I suspect 15 has a different module system; so maybe not a bug.

src/N3Store.js
dhurlburtusa: I'd be happy to make another PR and change all the nested functions to arrow functions and remove all the `var self = this;` uses and other unnecessary closure creations.

Here is list of some of the areas that would be updated.

- N3Lexer.js in tokenize
- N3Parser.js in parse
- N3Store in _uniqueEntities, import, and remove
- N3StreamParser constructor and import
- N3StreamWriter constructor and import

There may be others too.

RubenVerborgh: @dhurlburtusa Thanks, that would be nice indeed! Could you change this single one already for this PR? Then I will merge.

dhurlburtusa: Updated.

RubenVerborgh:
Rebased to circumvent the Node 15 issue (#229).

PR #226, opened 2020-10-28
Switch to arrow functions

### LINKED ISSUE #225: Switch to Using Arrow Functions to Minimize Explicit Closure Creation
Take advantage of the simplicity that arrow functions bring.

dhurlburtusa:
Closes #225.

Note: I stacked this branch on top of the patch-1 branch from the #223 PR.

I went ahead and changed several nested functions that were not requiring access to `this` too. Please let me know if you intended for those to remain unchanged.

Note: I did not update the nested functions in the test suite. Last time I used mocha, it was recommended not to use arrow functions. I am not sure if that is still the case. Please let me know if you would like the test suite updated to use arrow functions.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -121,6 +121,7 @@
 
     // Stylistic Issues
     array-bracket-spacing: 2,
+    arrow-parens: [2, "as-needed"],
     block-spacing: 2,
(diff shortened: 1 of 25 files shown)


RubenVerborgh:
Thanks so much @dhurlburtusa!

PR #231, opened 2020-11-12
Allow using import statements in Node

### LINKED ISSUE #196: Export ESM as well

Vinnl:
Although the module is still exported as a CommonJS module, this
allows developers to import it using an import statement when
writing Node code in ES modules.

Fixes #196.

For more background and discussion, see https://github.com/rdfjs/N3.js/issues/196#issuecomment-726173831.

### DIFF
--- a/package.json
+++ b/package.json
@@ -13,6 +13,10 @@
   "license": "MIT",
   "main": "./lib/index.js",
   "module": "./src/index.js",
+  "exports": {
+    "require": "./lib/index.js",
+    "import": "./lib/index.js"
+  },
   "sideEffects": false,
   "engines": {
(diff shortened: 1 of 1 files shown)


Vinnl:
Actually, looks like I got things mixed up again. This isn't necessary to make things work in Node+ES modules, and doesn't solve the lack of a default export in bundlers. Dammit, stupid module systems.

RubenVerborgh:
Whatever we do, let's have tests as in https://github.com/RubenVerborgh/AsyncIterator/blob/v3.0.3/.travis.yml#L25-L34 so we are sure it works and keeps on working in different Node versions.

PR #232, opened 2020-11-12
Allow using import statements in Node

### LINKED ISSUE #196: Export ESM as well

Vinnl:
Although the module is still exported as a CommonJS module, this
allows developers to import it using an import statement when
writing Node code in ES modules.

This is pretty risk-free: it only _adds_ an additional export. It does make the default export explicitly part of the API, but that is actually already the case, i.e. if you import from n3 in an ES module in Node.

This fixes #196, at least in the sense that its title says "provide a default export".

### DIFF
--- a/package.json
+++ b/package.json
@@ -13,6 +13,10 @@
   "license": "MIT",
   "main": "./lib/index.js",
   "module": "./src/index.js",
+  "exports": {
+    "require": "./lib/index.js",
+    "import": "./lib/index.js"
+  },
   "sideEffects": false,
   "engines": {
     "node": ">=8.0"


Vinnl:
(Hmm, the test that's failing in Travis succeeds for me on Node 14...)

RubenVerborgh:
Node 14 has an unrelated problem, so no worries.

RubenVerborgh:
Will need tests though as per https://github.com/rdfjs/N3.js/pull/231#issuecomment-726181564

Vinnl:
Unfortunately I don't have the bandwidth to set up the build infrastructure on a build system I don't have access to to test imports in different environments, but as mentioned in https://github.com/rdfjs/N3.js/issues/196#issuecomment-726188276, I'd judge this addition to be safe enough to not need such tests (since it doesn't alter the module system - just like there are no tests for the existing exports). If someone were to go all the way to set up that infrastructure, I'd go all the way and properly add ES modules as described in https://github.com/rdfjs/N3.js/issues/196#issuecomment-726173831.

If you think this does need such tests, and prefer to go without isomorphic Node-ESM+Webpack support until someone has time to set all that up, then feel free to close this - maybe it'll be helpful to someone in the future picking it up.

RubenVerborgh:
> If you think this does need such tests

I'm afraid so; the module system is really tricky, and I can only merge if we're sure it works.
After all, this PR is the exact same as the closed #231.

No worries if you don't have the bandwidth, but then we better wait indeed.

Vinnl:
@RubenVerborgh Oh whoops, thanks for mentioning that! I selected the wrong branch >.< That must've been a weird conversation for you... See #233.

RubenVerborgh:
Haha, thanks 🙂

PR #233, opened 2020-11-12
Add a default export

### LINKED ISSUE #196: Export ESM as well

Vinnl:
When importing n3 from an ES module in Node, the exports will be
treated as a default export, because the package itself is not an
ES module.

This _actually_ exports that default module, so that the same
syntax (i.e. importing the default export) can be used with both
Node and bundlers.

(Replaces #232 - this is what I intended to submit there.)

Fixes #196.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -22,7 +22,33 @@ import {
   termToId,
 } from './N3DataFactory';
 
+// Named exports
 export {
+  Lexer,
+  Parser,
+  Writer,
+  Store,
+  StreamParser,
+  StreamWriter,
+  Util,
+
+  DataFactory,
+
+  Term,
+  NamedNode,
+  Literal,
+  BlankNode,
+  Variable,
+  DefaultGraph,
+  Quad,
+  Triple,
+
+  termFromId,
+  termToId,
+};
+
+// Export all named exports as a default object for backward compatibility
+export default {
   Lexer,
   Parser,
   Writer,


eweitz:
Naive question: would this let me seamlessly `import {foo} from 'n3'` in a JS library built for a browser context?

I'm evaluating SPARQL packages for use in [a JavaScript web component](https://github.com/eweitz/ideogram).  I'd like to be able to declare `n3` as a dependency in my `package.json`, and have webpack bundle it like all my other dependencies.  So the README's line about [needing to create a UMD bundle](https://github.com/rdfjs/N3.js#installation) gives me pause -- to me it suggests I'd need special handling for this one dependency.

It seems this PR would resolve my concern.  Is that right?

Vinnl:
@eweitz You can just give it a try :) But if you use Webpack you should be able to `import { foo } from "n3"` even without this PR. This PR solves the issue where you can't import from n3 if you're building something that should work in _both_ Node and the browser (or at least, not without [a cumbersome workaround](https://github.com/inrupt/solid-client-js/blob/2c4cfb891b70ffa64ae7b2f5965a12f2cb0ce5c7/src/formats/turtle.ts#L91-L113)).

eweitz:
Thanks!  That was just the clarification I needed.

Vinnl:
Yeah as far as I am concerned, although there were build failures for reasons that I'm not sure are related to this PR.

RubenVerborgh:
@Vinnl Thanks. I rebased, let's see.

RubenVerborgh:
Looks good!

RubenVerborgh (approved):

RubenVerborgh:
> so that the same
> syntax (i.e. importing the default export) can be used with both
> Node and bundlers.

Although… this would kill tree-shaking, wouldn't it?
Isn't the opposite what we want?

Vinnl:
@RubenVerborgh Yes, it would kill tree shaking _for people who use the default export_. If you're building something browser-based, you can just continue to use the named exports. If you're building something that runs in both, you'd have to use the default export, which does not work with tree shaking - but the alternative is not using n3 at all :)

(The cumbersome workaround I mentioned above also kills tree-shaking.)

In the longer term, you would want to publish n3 as an ES module in Node as well (i.e. with `"type": "module"`), but that also requires careful fiddling with package exports to make sure it continues to work with CommonJS projects in Node.

RubenVerborgh:
> In the longer term, you would want to publish n3 as an ES module in Node as well (i.e. with `"type": "module"`)

Thanks. I'm gonna do that. Don't want people to start depending on the default export.

Vinnl:
Keep in mind that people using Node 12.7 and above are already depending on the default export. Setting `"type": "module"` is a breaking change, unless you preserve that.

RubenVerborgh:
> Keep in mind that people using Node 12.7 and above are already depending on the default export.

Thanks, good point. Ugh. Guess I'll need both changes then.
And good tests.

Vinnl:
In case it's helpful, we've got [a number of such tests](https://github.com/inrupt/solid-client-js/blob/2c4cfb891b70ffa64ae7b2f5965a12f2cb0ce5c7/.github/workflows/cd.yml#L155-L257) for solid-client as well (using [these minimal projects](https://github.com/inrupt/solid-client-js/tree/main/.github/workflows/cd-packaging-tests) - alternatively, [here](https://github.com/inrupt/jose-legacy-modules-js/blob/253e524e287454665a31b52ece7f14a0063cd468/.github/workflows/publish.yml#L66-L144)'s another project that runs the tests without publishing to npm first).

RubenVerborgh:
> we've got [a number of such tests](https://github.com/inrupt/solid-client-js/blob/2c4cfb891b70ffa64ae7b2f5965a12f2cb0ce5c7/.github/workflows/cd.yml#L155-L257) for solid-client as well

Ooh nice, good candidates for packaging up into GitHub Actions! Will definitely be useful.

RubenVerborgh:
Will track in https://github.com/rdfjs/N3.js/issues/196.

PR #234, opened 2020-12-11
Added Architecure ppc64le

genisysram:
Hi
Here is my contribution to your code, its working good on powersystems.

Thanks for the code, its working good.
## What do these changes do?
Added Architecture "ppc64le"
## Are there changes in behavior for the user?
No

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,7 @@
 sudo: false
+arch:
+    - amd64
+    - ppc64le
 language: node_js
 node_js:
  -  8


RubenVerborgh:
Thanks for the PR. However, there is no code in here that is specific to a certain CPU architecture, so I prefer to not multiply the number of needed platforms in order to practice green computing 🌲

genisysram:
we have ran this code in PowerSystems by adding arch:ppc64le , rigorously tested and ensures it working perfectly there

RubenVerborgh:
Is there any reason why the code _wouldn't_ run on any specific CPU architecture, provided Node.js has been ported and tested?

PR #240, opened 2021-02-16
Make N3Store Implement RDF/JS DatasetCore Interface

### LINKED ISSUE #237: Add RDF.js Dataset interface to N3.Store?
Some Javascript libraries (like rdf-validate-shacl) now accept and return RDF.js Datasets but I don't think they will work with N3.Store. I am wondering about adding the Dataset interface either to the N3.Store or a wrapper that would make it look like a Dataset.

Have you already decided that is a bad idea? Are you interested in including that?

dhurlburtusa:
Closes #237.

### DIFF
--- a/src/N3Store.js
+++ b/src/N3Store.js
@@ -4,7 +4,7 @@ import { Readable } from 'readable-stream';
 import namespaces from './IRIs';
 
 // ## Constructor
-export default class N3Store {
+class N3Store {
   constructor(quads, options) {
     // The number of quads is initially zero
     this._size = 0;
@@ -211,6 +211,14 @@ export default class N3Store {
 
   // ## Public methods
 
+  // ### `add` adds the specified quad to the dataset.
+  // Returns the dataset instance it was called on.
+  // Existing quads, as defined in Quad.equals, will be ignored.
+  add(quad) {
+    this.addQuad(quad);
+    return this;
+  }
+
   // ### `addQuad` adds a new quad to the store.
   // Returns if the quad index has changed, if the quad did not already exist.
   addQuad(subject, predicate, object, graph) {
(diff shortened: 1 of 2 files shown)


src/N3Store.js
```diff
@@ -211,6 +211,14 @@ export default class N3Store {
 
   // ## Public methods
 
+  // ### `add` adds the specified quad to the dataset.
```
dhurlburtusa, line 214: For the new methods, I copied the documentation from the RDF/JS DatasetCore spec and tweaked it a bit. Please let me know if there is anything you want me to change.

src/N3Store.js
```diff
@@ -316,7 +342,15 @@ export default class N3Store {
   // ### `removeMatches` removes all matching quads from the store
   // Setting any field to `undefined` or `null` indicates a wildcard.
   removeMatches(subject, predicate, object, graph) {
-    return this.remove(this.match(subject, predicate, object, graph));
+    const stream = new Readable({ objectMode: true });
```
dhurlburtusa, line 345: Since the old implementation of this.match was only being used here, I just inlined the old implementation.

src/N3Store.js
```diff
@@ -790,9 +820,139 @@ export default class N3Store {
...
 // Determines whether the argument is a string
 function isString(s) {
   return typeof s === 'string' || s instanceof String;
 }
+
+/**
+ * A class that implements both the DatasetCore interface and the Readable stream
+ * interface.
+ */
+class DatasetCoreAndReadableStream {
```
dhurlburtusa, line 841: I'm open to suggestions to improve the implementation. I believe I captured all possible public getters and methods of the Readable stream interface.

dhurlburtusa, line 841: Another way to implement this functionality is to instead of returning a hybrid object from match is to make the N3Store itself implement the Readable stream interface and return an instance of N3Store from match. That way, users can treat it like a Readable stream, a DatasetCore, or the full N3Store. Otherwise, only the 4 methods of DatasetCore are available plus the ones for Readable stream. Thoughts?

RubenVerborgh:
Thanks, @dhurlburtusa. Currently swamped, but will look into this.

RubenVerborgh:
Thanks for the implementation and your patience.

Reviewed and merged as cf8d28f0c05fc327151c1a76e333f2f8f0fce819.

PR #241, opened 2021-02-24
[StreamWriter] Use highWaterMark of 16

LaurensRietveld:
The streamwriter currently uses a highWaterMark of 16.384, which is a value that is typically used for buffered streams. 
As a consequence, some streams may have atypical behaviour (in runtime, progress, and memory usage).

Considering this is an object stream (at least the writable part of it), I'd propose using a highWaterMark of 16 (the nodejs default)

### DIFF
--- a/src/N3StreamWriter.js
+++ b/src/N3StreamWriter.js
@@ -5,8 +5,7 @@ import N3Writer from './N3Writer';
 // ## Constructor
 export default class N3StreamWriter extends Transform {
   constructor(options) {
-    super({ encoding: 'utf8' });
-    this._writableState.objectMode = true;
+    super({ encoding: 'utf8', writableObjectMode: true });
 
     // Set up writer with a dummy stream object
     var self = this;


RubenVerborgh:
> I'd propose using a highWaterMark of 16 (the nodejs default)

But if it's a default, why do we need to set it?

LaurensRietveld:
> > I'd propose using a highWaterMark of 16 (the nodejs default)
> 
> But if it's a default, why do we need to set it?

See [here](https://nodejs.org/api/stream.html#stream_new_stream_writable_options) for the highwatermarks used by nodejs when running in object mode or buffered mode. 
We had to set this value, because we were instantiating the transform stream (the super() class) and modifying the objectmode property afterwards.

I've modified the PR in my latest commit, and took a slightly different approach. We're now passing the `objectMode` via the `super()` call. This means the transform stream will use the appropriate default highWaterMark of 16.

RubenVerborgh:
Thanks; that might explain some of the behavior we've seen with streams waiting longer than they needed to.

PR #249, opened 2021-05-12
Callback activation for prefix event

fvitali:
Added callback activation for prefix event (line 26). This is a required change for another Pull Request on rdf-parse


RubenVerborgh (changes requested):
Please explain the necessity.
Please add a test case.

PR #258, opened 2021-07-22
Lexer should ignore byte order mark at beginning of file. (closes #242)

### LINKED ISSUE #242: Ignore Byte-Order Mark at start
As suggested here https://github.com/comunica/comunica/issues/778#issuecomment-774482524

jitsedesmet:


src/N3Lexer.js
```diff
@@ -471,7 +478,14 @@ export default class N3Lexer {
...
           else {
-            this._input += data;
+            const dataAsString = String(data);
+            if (this._dataPrependTimes === 0) {
+              this._input = this._ignoreStartingBom(dataAsString);
+            }
+            else {
+              this._input += dataAsString;
+            }
+            this._dataPrependTimes++;
```
RubenVerborgh, line 488: Mmm can we do this more nicely? Like a boolean `this._bomMarkChecked` set by `_ignoreStartingBom`?

src/N3Lexer.js
```diff
@@ -429,6 +429,11 @@ export default class N3Lexer {
     return err;
   }
 
+  _ignoreStartingBom(input) {
```
RubenVerborgh, line 432:
```suggestion
  _readStartingBom(input) {
```

src/N3Lexer.js
```diff
@@ -429,6 +429,11 @@ export default class N3Lexer {
...
+    const bom = /^(\ufeff?)/.exec(input);
+    return input.substr(bom[1].length, input.length);
```
RubenVerborgh, line 434:
```suggestion
   if (input.startsWith('\ufeff')) …
```

to avoid unnecessarily subsetting (very large) strings.

src/N3Lexer.js
```diff
@@ -471,7 +478,14 @@ export default class N3Lexer {
           }
           // Otherwise, tokenize as far as possible
           else {
-            this._input += data;
+            const dataAsString = String(data);
+            if (this._dataPrependTimes === 0) {
```
RubenVerborgh, line 482: Note that we could have the pathological case that the first input chunk is an empty string, and then the BOM mark comes. So `this._bomMarkChecked` (see below) should only be set once _some_ first character has been read.

RubenVerborgh (changes requested):
Good start, I have some perfomance concerns (notes inline).

I wonder whether `\ufeff` can be checked for in the big `switch` statement, and only be consumed if the read position is 0. This to avoid checking for position zero every time we go into the main read loop.

src/N3Lexer.js
jitsedesmet: Yes, we could do that. I think that's a matter of taste. I'll change it to a boolean.

src/N3Lexer.js
jitsedesmet: Wow, hadn't thought about that. Thanks

src/N3Lexer.js
jitsedesmet: You're right, combined with all other suggestions, this makes a much prettier result.

jitsedesmet:
I don't understand what you mean with:

> I wonder whether \ufeff can be checked for in the big switch statement, and only be consumed if the read position is 0. This to avoid checking for position zero every time we go into the main read loop.

Is this a suggestion outside the once you made? Can you explain it some more?

src/N3Lexer.js
RubenVerborgh: It might seem insignificant, but this function is called millions of times per second. Any write we don't have to do, has impact. (Plus at some point it would overflow to Infinity, that limit is easily reached when parsing a couple of 100 million triples.)

src/N3Lexer.js
```diff
@@ -471,7 +485,7 @@ export default class N3Lexer {
           }
           // Otherwise, tokenize as far as possible
           else {
-            this._input += data;
+            this._input += this._readStartingBom(String(data));
```
rubensworks, line 488: Is this `String` constructor necessary?

jitsedesmet: Yes otherwise it will fail with `Uncaught TypeError: input.startsWith is not a function`.

RubenVerborgh: What type if `input` then if that happens?  It's supposed to be a string already. (Buffer?)

jitsedesmet: Apparently there are 7 tests where typeof input === Object

RubenVerborgh: I think `Buffer`. Will have a look myself. Could you just add a test case for `streamOf(['', ''\ufeff''])`? Then I'll finish up. Thanks!

jitsedesmet: Done. I fear I pushed changes to the package-lock.json file. Look out for that!

RubenVerborgh:
Thanks @jitsedesmet, I was able to simplify it further into f393b6c61078addbf974896e2e575c1c8e8cf6d9 by using `this._input` as a signal of whether parsing has started.

PR #260, opened 2021-10-12
Add CDN script to README

NoelDeMartin:
I just published [n3-browserify](https://www.npmjs.com/package/n3-browserify) to npm in order to allow using the library from a CDN. I haven't done a lot of testing, but at least I tried parsing Turtle strings and it works. Let me know if I'm doing something wrong.

You don't have to merge this PR if you prefer to keep this undocumented, but I just wanted to let the maintainers of the project know. I'm also happy to transfer ownership of the npm package if you prefer to manage this yourselves. Or maybe you prefer to include a browser bundle in your own npm package. In any case, this gets the job done for now :).

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,6 +48,12 @@ or
 [_Introduction to browserify_](https://writingjavascript.org/posts/introduction-to-browserify).
 You will need to create a "UMD bundle" and supply a name (e.g. with the `-s N3` option in browserify).
 
+You can also use N3.js from a CDN with the following script:
+
+```html
+<script src="https://cdn.jsdelivr.net/npm/n3-browserify@latest"></script>
+```
+
 ## Creating triples/quads
 N3.js follows the [RDF.js low-level specification](http://rdf.js.org/).
 


RubenVerborgh:
Thanks @NoelDeMartin! My main concern would be that versions get out of date; maybe this script should be part of the present repository instead?

NoelDeMartin:
@RubenVerborgh if you want, I can open a PR to add a new build script that creates this browserify bundle in something like `dist/index.js` (we can discuss the particulars later). Declaring that bundle with a `browser` entry in the `package.json` should allow using a cdn like this: https://cdn.jsdelivr.net/npm/n3 At the moment this works, but it's returning the build on `lib` which is intended for node.

RubenVerborgh:
Sounds good!

Is Browserify still the latest though? Might want to do a quick webpack or so.

NoelDeMartin:
> Is Browserify still the latest though? Might want to do a quick webpack or so.

I have no idea, I just used browserify because it was mentioned in the README and I had seen this type of `*-browserify` packages before. I'll take a look to what's better before making the PR :).

PR #261, opened 2021-11-05
Configure browser build

NoelDeMartin:
As a follow up to #260, I've configured the browser build. Using this approach, `n3-browserify` won't be necessary any longer and the library will work out of the box with a url such as `https://cdn.jsdelivr.net/npm/n3`.

I tried using webpack, but the latest version doesn't include node polyfills out of the box; we'd need to use webpack 4 or configure some polyfills. So I tried using rollup instead, but the polyfills I tried didn't work either and I decided to just use browserify. It works, it's simple enough, and I don't see a reason why we wouldn't use it (after all, it's one of the suggestions in the README).

I'm sure it is possible to use webpack or rollup, but I didn't manage to make it work and I don't want to spend any more time on this. If anyone else knows more about this and wants to give it a try, go ahead.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,7 @@
 *.log
 .nyc_output
 .rdf-test-suite-cache
+browser
 docs
 lib
 node_modules
--- a/README.md
+++ b/README.md
@@ -48,6 +48,11 @@ or
 [_Introduction to browserify_](https://writingjavascript.org/posts/introduction-to-browserify).
 You will need to create a "UMD bundle" and supply a name (e.g. with the `-s N3` option in browserify).
 
+You can also load it via CDN:
+```html
+<script src="https://unpkg.com/n3/browser/n3.min.js"></script>
+```
+
 ## Creating triples/quads
 N3.js follows the [RDF.js low-level specification](http://rdf.js.org/).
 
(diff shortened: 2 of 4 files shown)


rubensworks:
While I understand the need, downside of this change is that treeshaking for the browser apps using Node.js won't be possible anymore, as all dependencies have been packed up in the dist file. So this will lead to larger bundle sizes. (Mainly concerned about Comunica, because its bundle size is already quite big)

Perhaps ESM is a viable alternative?

NoelDeMartin:
@rubensworks The use-case that this is trying to solve is using CDN scripts, anything that relies on a bundling mechanism that supports tree-shaking should ignore this build and import the ESM build, configuring polyfills if necessary.

I thought that was already the case, but maybe by declaring this in the `browser` entry of the `package.json` it can be an issue for bundling as well? We could avoid declaring it in the `package.json` and a CDN link would still work, even though it'd be more verbose, something like this: `https://cdn.jsdelivr.net/npm/n3/dist/n3.js`.

rubensworks:
> I thought that was already the case, but maybe by declaring this in the browser entry of the package.json it can be an issue for bundling as well?

Indeed, that was what I intended to refer to.

> We could avoid declaring it in the package.json and a CDN link would still work, even though it'd be more verbose, something like this: https://cdn.jsdelivr.net/npm/n3/dist/n3.js.

That would be a good solution IMO.

NoelDeMartin:
Ok I just updated the commit to remove the `browser` declaration from `package.json`.

Since it's won't be the default url, I added the instructions to the README. I used `unpkg.com` because that's what Vue uses in their documentation, let me know if you prefer another CDN or removing it from the README altogether.

RubenVerborgh:
@NoelDeMartin Pardon the delay; I made minor changes (using the `browser` instead of `dist` folder). This okay with you?

RubenVerborgh (approved):

NoelDeMartin:
Thanks @RubenVerborgh, yes it looks good!

Issue #262, opened 2021-11-06
Notation3 parsing : no support for blank nodes as predicate?


phochste:
In version 1.11.1 I get a parsing error when reading:

```
@prefix : <http://example.org/ex#>.
:Subject  [ a :Predicate ] :Object.
```

Got a

```
Error: Expected entity but got [ on line 3.
```

RubenVerborgh:
Thanks @phochste, would you be available to test https://github.com/rdfjs/N3.js/pull/263?

phochste:
> Thanks @phochste, would you be available to test #263?

Yes!

PR #263, opened 2021-11-08
Allow blank nodes as predicates in N3.

RubenVerborgh:
Fixes https://github.com/rdfjs/N3.js/issues/262


phochste (approved):
Works for me!

PR #270, opened 2022-01-19
Export ES6 module by default

RubenVerborgh:


RubenVerborgh:
Well, not today then.

PR #271, opened 2022-01-28
Support passing NamedNode to prefix function (fixes #247)

### LINKED ISSUE #247: Passing NamedNode to prefix function throws error
The following lines throws an error:

```
var N3 = require('n3');
N3.Util.prefix(N3.DataFactory.namedNode('a:a'))
```
The thrown error is:
```
Error: Unknown prefix: 
    at processPrefix (/home/lrd900/triply/N3.js/lib/N3Util.js:78:13)
    at prefixes (/home/lrd900/triply/N3.js/lib/N3Util.js:62:41)
    at Object.prefix (/home/lrd900/triply/N3.js/lib/N3Util.js:52:10)
    at Object.<anonymous> (/home/lrd900/triply/N3.js/test.js:3:9)
```

I'm a bit confused over whether this is a bug or not. The documentation does not specify whether this is supported or not. The typescript interface does explicitly allow passing NamedNodes.

(my 2c: I'd like to support passing NamedNodes here)

MPvHarmelen:


src/N3Util.js
```diff
@@ -34,6 +34,7 @@ export function inDefaultGraph(quad) {
 
 // Creates a function that prepends the given IRI to a local name
 export function prefix(iri, factory) {
+  if (isNamedNode(iri)) iri = iri.value;
   return prefixes({ '': iri }, factory)('');
```
RubenVerborgh, line 38: Might as well support literals too?
```suggestion
  return prefixes({ '': iri.value || iri }, factory)('');
```

test/N3Util-test.js
```diff
@@ -196,6 +196,16 @@ describe('Util', () => {
         expect(rdfs('label')).to.deep.equal(new NamedNode('http://www.w3.org/2000/01/rdf-schema#label'));
       });
 
+      it('should expand a NamedNode prefix', () => {
+        const rdfs = Util.prefix(new NamedNode('http://www.w3.org/2000/01/rdf-schema#'));
+        expect(rdfs('label')).to.deep.equal(new NamedNode('http://www.w3.org/2000/01/rdf-schema#label'));
+      });
+
+      it('should expand an empty literal prefix', () => {
+        const empty = Util.prefix(new Literal(''));
```
RubenVerborgh, line 205: Well, rather a literal with a URL in there… let's hope they never use it for this 😄

RubenVerborgh (commented):
Almost good, just fix that test. Could you please squash all commits?

test/N3Util-test.js
MPvHarmelen: I added this test because it breaks when using `||` but doesn't when using `??`. But I definitely agree this isn't the intended use of `Util.prefix` :sweat_smile:

RubenVerborgh:
Thanks, merged as e37b4df3c07bbfcb28134341f2d2ef14513b1f1f.

PR #273, opened 2022-03-19
Add start and end index to tokens

alxflam:
This PR adds the start and end index to the token object.

I would like to have the start and end index available as a N3.js consumer, because that helps to locate the exact location of parser errors. In my case, i'm using N3.js plugged into a Language Server and i want to create error markers in the editor for all parser errors - currently i always just add them add the beginning of the line as only the line information is present for tokens.

Btw, I did not adapt tests yet as i first wanted to get some feedback whether it's okay to add the start and end indices.

Greetings,
Alex

### DIFF
--- a/src/N3Lexer.js
+++ b/src/N3Lexer.js
@@ -85,7 +85,7 @@ export default class N3Lexer {
       while (whiteSpaceMatch = this._newline.exec(input)) {
         // Try to find a comment
         if (this._comments && (comment = this._comment.exec(whiteSpaceMatch[0])))
-          callback(null, { line: this._line, type: 'comment', value: comment[1], prefix: '' });
+          callback(null, this._createToken(this._line, 'comment', comment[1], '', currentLineLength - input.length, currentLineLength - input.length + whiteSpaceMatch[0].length));
         // Advance the input
         input = input.substr(whiteSpaceMatch[0].length, input.length);
         currentLineLength = input.length;
@@ -101,8 +101,8 @@ export default class N3Lexer {
         if (inputFinished) {
           // Try to find a final comment
           if (this._comments && (comment = this._comment.exec(input)))
-            callback(null, { line: this._line, type: 'comment', value: comment[1], prefix: '' });
-          callback(input = null, { line: this._line, type: 'eof', value: '', prefix: '' });
+            callback(null, this._createToken(this._line, 'comment', comment[1], '', currentLineLength - input.length, currentLineLength));
+          callback(input = null, this._createToken(this._line, 'eof', '', '', currentLineLength, currentLineLength));
         }
         return this._input = input;
       }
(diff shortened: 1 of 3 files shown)


RubenVerborgh:
The main thing we'll need is a detailed performance assessment in terms of speed and memory usage.

alxflam:
I I have made some performance measurements with the following test data:
Dataset 1: https://databus.dbpedia.org/dbpedia/generic/homepages/2021.12.01/homepages_lang=en.ttl.bz2
(2.925.341 tokens)
Dataset 2: https://www.nlm.nih.gov/databases/download/mesh.html (66.576.035 tokens)
I ran the lexer performance script 5 times for each of these datasets.

For Dataset 1, my result is:
Unchanged: Average of 1,1756s runtime and 68,2MB memory
With Index: Average of 1,1628s runtime and 68,2MB memory

Dataset 2:
Unchanged: Average of 22,085s runtime and 70,67MB memory
With Index: Average of 21,8045s runtime and 71,6MB memory

Hence i'd conclude there's no performance impact and only an increased memory consumption (but only noticeable for very large datasets). I was curious because the index change somehow led to a tiny bit better performance, but if i run the tests afterwards again 5 times each (with and without index), then sometimes the result is the other way round (but still the performance difference is around the same: around 0,015s for dataset 1 and about 0,25s for dataset 2).

@RubenVerborgh what do you think?

RubenVerborgh:
Looks great, thanks. I can accept.
For the tests, it's fine if the indexes are only tested in a small number of places.

alxflam:
@RubenVerborgh i adapted #shouldTokenize in the lexer test to ignore the start and end index (ignored as long as they are not present in the expected token given) and added separate tests for the index handling. 

This is how i can use it to position markers:
parseErrorIndex

Notice that the token range (start to end index) includes the whitespaces - i think this is correct, as that's the range that got processed and resulted in the token. And for users of the parser it's now easy to calculate a different end index by checking for trailing whitespaces of a token range. Otherwise the lexer would need to calculate the "trimmed" end index.

RubenVerborgh:
Looking good indeed!

RubenVerborgh:
I reviewed at https://github.com/rdfjs/N3.js/commit/fa4179558ba881c071cce49f31ec232f79a51b9e (could not push to your branch).

A remaining issue is that `callback` is still called on lines 88, 104, 105 with a token that has a different form. On the one hand, this is a matter of consistency for developers, that they can always expect the `start` and `end` fields to be part of a token. On the other hand, this is a matter of monomorphism for the V8 engine, to make sure that the argument has the same shape so it can continue to be optimized.

Perhaps we need a small helper function; either as a member function or an inline function, to consistently output the token.

alxflam:
@RubenVerborgh good point. As start and end index determination is different for the four token creation locations, i have only created a utility method to create the token which receives all necessary parameters. Added a further test dedicated to comments and eof wrt. start and end index. You should have received a GitHub invitation now so you can directly push to my fork.

RubenVerborgh (approved):

RubenVerborgh:
Thanks a lot! Available as v1.15.0.

RubenVerborgh:
PS Can you link to your language server when ready?

alxflam:
@RubenVerborgh sure, i'll do.

PR #285, opened 2022-04-13
chore: use readquads in filtered (to improve perf.)

jeswr:
This skips the unnecessary step of materializing quads in an array prior to loading them into the new `N3Store`.


jeswr:
Pinging @RubenVerborgh for review

RubenVerborgh (approved):

PR #290, opened 2022-04-30
chore: add tests for concurrent read/write behaviour

jeswr:
This adds some sanity checks for reading from a `match` iterator whilst concurrently adding new data, in light of the changes in v1.16 where we are now iterating over the index rather than eagerly placing all matches into a list.

### DIFF
--- a/test/N3Store-test.js
+++ b/test/N3Store-test.js
@@ -7,7 +7,7 @@ import {
   termFromId, termToId,
 } from '../src/';
 import namespaces from '../src/IRIs';
-import chai from 'chai';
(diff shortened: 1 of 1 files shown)


jeswr:
A use case for such behaviour, is in reasoning where we are adding new implicit facts whilst reading from the store at the same time to calculate new inferences.

The comunica reasoning components that I am working on will implicitly do this if the destination for the implicit data is set to be the same as the source of the raw data.

jeswr:
Pinging @RubenVerborgh for review

PR #296, opened 2022-05-13
N3 Reasoner Class

jeswr:
This introduces an N3 Reasoner class as discussed in #295 

For the record here is the output of the performance test, noting that my machine has ~20% variance in results as noted in past PRs
```
jeswr@debian:~/semi/N3.js$ NODE_ENV=production node --max-old-space-size=8192 perf/N3Reasoner-perf.js 
Reasoning over TimBL profile and FOAF
loading foaf ontology: 15.945ms
loading tim berners lee profile card: 13.481ms
Reasoning: 48.471ms

Running Deep Taxonomy Benchmark
Reasoning: test-dl-10.n3: 0.231ms
Reasoning: test-dl-100.n3: 1.019ms
Reasoning: test-dl-1000.n3: 26.149ms
Reasoning: test-dl-10000.n3: 76.392ms
Reasoning: test-dl-100000.n3: 1.035s
Reasoning: test-dl-1000000.n3: 10.606s
```

### DIFF
--- a/perf/data/foaf.ttl
+++ b/perf/data/foaf.ttl
@@ -1,0 +1,699 @@
+@prefix foaf: <http://xmlns.com/foaf/0.1/> .
+@prefix owl: <http://www.w3.org/2002/07/owl#> .
+@prefix dc11: <http://purl.org/dc/elements/1.1/> .
+@prefix wot: <http://xmlns.com/wot/0.1/> .
+@prefix rdfs: <http://www.w3.org/2000/01/rdf-schema#> .
+@prefix ns0: <http://www.w3.org/2003/06/sw-vocab-status/ns#> .
+@prefix schema: <http://schema.org/> .
+@prefix geo: <http://www.w3.org/2003/01/geo/wgs84_pos#> .
+@prefix dc: <http://purl.org/dc/terms/> .
+@prefix rdf: <http://www.w3.org/1999/02/22-rdf-syntax-ns#> .
+@prefix skos: <http://www.w3.org/2004/02/skos/core#> .
+
+foaf:
+  a owl:Ontology ;
+  dc11:title "Friend of a Friend (FOAF) vocabulary" ;
+  dc11:description "The Friend of a Friend (FOAF) RDF vocabulary, described using W3C RDF Schema and the Web Ontology Language." .
+
+wot:assurance a owl:AnnotationProperty .
+wot:src_assurance a owl:AnnotationProperty .
+<http://www.w3.org/2003/06/sw-vocab-status/ns#term_status> a owl:AnnotationProperty .
+dc11:description a owl:AnnotationProperty .
+dc11:title a owl:AnnotationProperty .
+dc11:date a owl:AnnotationProperty .
+rdfs:Class a owl:Class .
+foaf:LabelProperty
+  a rdfs:Class, owl:Class ;
+  ns0:term_status "unstable" ;
+  rdfs:label "Label Property" ;
+  rdfs:comment "A foaf:LabelProperty is any RDF property with texual values that serve as labels." ;
+  rdfs:isDefinedBy foaf: .
+
+foaf:Person
+  a rdfs:Class, owl:Class ;
+  rdfs:label "Person" ;
+  rdfs:comment "A person." ;
+  ns0:term_status "stable" ;
+  owl:equivalentClass schema:Person, <http://www.w3.org/2000/10/swap/pim/contact#Person> ;
+  rdfs:subClassOf foaf:Agent, geo:SpatialThing ;
+  rdfs:isDefinedBy foaf: ;
+  owl:disjointWith foaf:Organization, foaf:Project .
+
+foaf:Agent
+  a owl:Class, rdfs:Class ;
+  ns0:term_status "stable" ;
+  rdfs:label "Agent" ;
+  rdfs:comment "An agent (eg. person, group, software or physical artifact)." ;
+  owl:equivalentClass dc:Agent .
+
+geo:SpatialThing
+  a owl:Class ;
+  rdfs:label "Spatial Thing" .
+
+foaf:Document
+  a rdfs:Class, owl:Class ;
+  rdfs:label "Document" ;
+  rdfs:comment "A document." ;
+  ns0:term_status "stable" ;
+  owl:equivalentClass schema:CreativeWork ;
+  rdfs:isDefinedBy foaf: ;
+  owl:disjointWith foaf:Organization, foaf:Project .
+
+foaf:Organization
+  a rdfs:Class, owl:Class ;
+  rdfs:label "Organization" ;
+  rdfs:comment "An organization." ;
+  ns0:term_status "stable" ;
+  rdfs:subClassOf foaf:Agent ;
+  rdfs:isDefinedBy foaf: ;
+  owl:disjointWith foaf:Person, foaf:Document .
+
+foaf:Group
+  a rdfs:Class, owl:Class ;
+  ns0:term_status "stable" ;
+  rdfs:label "Group" ;
+  rdfs:comment "A class of Agents." ;
+  rdfs:subClassOf foaf:Agent .
+
+foaf:Project
+  a rdfs:Class, owl:Class ;
+  ns0:term_status "testing" ;
+  rdfs:label "Project" ;
+  rdfs:comment "A project (a collective endeavour of some kind)." ;
+  rdfs:isDefinedBy foaf: ;
+  owl:disjointWith foaf:Person, foaf:Document .
+
+foaf:Image
+  a rdfs:Class, owl:Class ;
+  ns0:term_status "stable" ;
+  rdfs:label "Image" ;
+  rdfs:comment "An image." ;
+  owl:equivalentClass schema:ImageObject ;
+  rdfs:subClassOf foaf:Document ;
+  rdfs:isDefinedBy foaf: .
+
+foaf:PersonalProfileDocument
+  a rdfs:Class, owl:Class ;
+  rdfs:label "PersonalProfileDocument" ;
+  rdfs:comment "A personal profile RDF document." ;
+  ns0:term_status "testing" ;
+  rdfs:subClassOf foaf:Document .
+
+foaf:OnlineAccount
+  a rdfs:Class, owl:Class ;
+  ns0:term_status "testing" ;
+  rdfs:label "Online Account" ;
+  rdfs:comment "An online account." ;
+  rdfs:isDefinedBy foaf: ;
+  rdfs:subClassOf owl:Thing .
+
+owl:Thing rdfs:label "Thing" .
+foaf:OnlineGamingAccount
+  a rdfs:Class, owl:Class ;
+  ns0:term_status "unstable" ;
+  rdfs:label "Online Gaming Account" ;
+  rdfs:comment "An online gaming account." ;
+  rdfs:subClassOf foaf:OnlineAccount ;
+  rdfs:isDefinedBy foaf: .
+
+foaf:OnlineEcommerceAccount
+  a rdfs:Class, owl:Class ;
+  ns0:term_status "unstable" ;
+  rdfs:label "Online E-commerce Account" ;
+  rdfs:comment "An online e-commerce account." ;
+  rdfs:subClassOf foaf:OnlineAccount ;
+  rdfs:isDefinedBy foaf: .
+
+foaf:OnlineChatAccount
+  a rdfs:Class, owl:Class ;
+  ns0:term_status "unstable" ;
+  rdfs:label "Online Chat Account" ;
+  rdfs:comment "An online chat account." ;
+  rdfs:subClassOf foaf:OnlineAccount ;
+  rdfs:isDefinedBy foaf: .
+
+foaf:mbox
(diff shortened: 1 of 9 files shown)


jeswr:
I just wanted to jot a few top-of-mind notes around a couple of modifications that should be made down the line to improve the features & performance of reasoning.

I'm not suggesting we do this now.


 - Track which quads are implicit vs. explicit. There are 2 ways I can immediately think of to do this
   1. In each dictionary use `true`/`false` as the value to indicate implicit/explicit rather than using `null`. This is my preference - but it would result in a slowdown if you are only trying to look up explicit, or only trying to look up implicit quads.
   2. Having 2 separate sets of indexes, 1 for implicit facts, the other for explicit. This is more likely to allow duplication of data (with explicit facts likely also to turn up in the implicit indices), will result in slower reasoning times due to having to look over 2 sets of indices, will result in slower lookup times for *all* data, again because you have to look over 2 sets of indices.
 - Handle incremental reasoning. There are several facets to this
   1. Allow the loading and preservation of rules
   2. The option to enable/disable incremental reasoning by default (i.e. should incremental reasoning be run every time `add`/`delete` is run). Though this has the potential for users to necessarily cause unnecessary slowdowns if they choose to repeatedly call `delete` on single quads rather than batching quads due to the nature of incremental maintenance algorithms
   3. Have an attribute to indicate whether the current database is consistent with the rules and explicit data.

src/N3Reasoner.js
```diff
@@ -0,0 +1,194 @@
...
+  // TODO [FUTURE]: Improve performance by 'pre-computing' the index lookup,
+  // e.g., if a rule only has one variable, then we can just give it a pointer
+  // to the index that it should do lookups from.
+  // Similarly with insertions.
+  _evaluatePremise(rule, content, cb, i = 0) {
+    let v1, v2, value, index1, index2;
+    const [val0, val1, val2] = rule.premise[i].value, index = content[rule.premise[i].content];
+    const v0 = !(value = val0.value);
+    for (value in v0 ? index : { [value]: index[value] }) {
+      if (index1 = index[value]) {
```
jeswr, line 26: Note for self when I get around to finish cleaning this up. Due to the way that N3.js handles deletions; this is only here for the case where value is *defined* and !(value in index). So we can reduce a bunch of unnecessary calcs by moving this check outside of the loop.

One option is
```ts
for (value in v0 ? index : value in index ? { [value]: index[value] } : {})
   ...
```
alternatively

phochste:
@jeswr I would love to see this code in N3. I have immediate plans to use this in some webapps I have in mind. 

Do you have in the reasoner support for native N3 lists? In my own attempts I've struggled with that. Dörthe Arndt created a test case for me a while ago. How would this be expressed in the N3 Reasoner?

```
PREFIX : <http://example.org/socrates#>

:Socrates a :Man.
:Man rdfs:subClassOf :Mortal.
 
_:x a :Man.
 
{
    ?A rdfs:subClassOf ?B. 
    ?S a ?A.
} => 
{
    ?S a ?B
}.
 
{ 
  ?x a ?y. 
  ?x a ?z.
}
=>
{
 ?x :aa (?y ?z)
}.
```

See also: http://ppr.cs.dal.ca:3002/n3/editor/s/C688Vckv

jeswr:
> I would love to see this code in N3.

As would I, however, I am currently undecided on whether the API should be as it is at present, or, whether `N3Reasoner` should extend `N3Store` and supply a `.reason` method. I think for the sake of supporting things like incremental reasoning (as mentioned [here](https://github.com/rdfjs/N3.js/pull/296#issuecomment-1126698258)) it should actually be the latter (@rubensworks do you have a view on this given your considerable work on the RDFJS spec?).

That said it shouldn't be much work for you to change between the two usages so you could always just use the branch in its current state in your app by running `npm i https://github.com/jeswr/N3.js/tree/feat/N3Reasoner` or `npm i https://github.com/rdfjs/N3.js/tree/a13e50e20c18eb0bc869db2a7c09fe552c653a4e`

> Do you have in the reasoner support for native N3 lists?

Not in the current state of this PR, as that would require the capability to be able to mint new blank nodes in order to represent the list.

Adding this capability shouldn't be *too* difficult, but I would prefer to look into how other reasoners do it and think about the best way of encoding that in rules so as to not limit ourselves from future performance improvements.

phochste:
@jeswr thanks this works. I see that the `reason` method inline updates the N3Store. Would it be possible to also support that the reason method returns a new N3Store while keeping the old one intact.

```
  let newStore = N3.Reasoner(oldStore).reason([ { premise:... , conclusion: ... } ]);
```

rubensworks:
> As would I, however, I am currently undecided on whether the API should be as it is at present, or, whether N3Reasoner should extend N3Store and supply a .reason method. I think for the sake of supporting things like incremental reasoning (as mentioned https://github.com/rdfjs/N3.js/pull/296#issuecomment-1126698258) it should actually be the latter (@rubensworks do you have a view on this given your considerable work on the RDFJS spec?).

+1 on `N3Reasoner`. However, instead of making this an extension of `N3Store`, it might be more interesting to use composition for this, as the latter is more flexible. This would also move more in the direction of making this reasoning logic more generic at RDF/JS-level in the future.

perf/data/foaf.ttl
```diff
@@ -0,0 +1,699 @@
...
+  rdfs:label "Online Chat Account" ;
+  rdfs:comment "An online chat account." ;
+  rdfs:subClassOf foaf:OnlineAccount ;
+  rdfs:isDefinedBy foaf: .
+
+foaf:mbox
+  a rdf:Property, owl:InverseFunctionalProperty, owl:ObjectProperty ;
+  ns0:term_status "stable" ;
+  rdfs:label "personal mailbox" ;
+  rdfs:comment "A  personal mailbox, ie. an Internet mailbox associated with exactly one owner, the first owner of this mailbox. This is a 'static inverse functional property', in that  there is (across time and change) at most one individual that ever has any particular value for foaf:mbox." ;
```
TallTed, line 139:
```suggestion
  rdfs:comment "A personal mailbox, i.e., an Internet mailbox associated with exactly one owner, the first owner of this mailbox. This is a 'static inverse functional property', in that there is (across time and change) at most one individual that ever has any particular value for foaf:mbox." ;
```

perf/data/foaf.ttl
```diff
@@ -0,0 +1,699 @@
...
+  ns0:term_status "testing" ;
+  rdfs:label "title" ;
+  rdfs:comment "Title (Mr, Mrs, Ms, Dr. etc)" ;
+  rdfs:isDefinedBy foaf: .
+
+foaf:nick
+  a rdf:Property, owl:DatatypeProperty ;
+  ns0:term_status "testing" ;
+  rdfs:label "nickname" ;
+  rdfs:comment "A short informal nickname characterising an agent (includes login identifiers, IRC and other chat nicknames)." ;
```
TallTed, line 207:
```suggestion
  rdfs:comment "A short informal nickname characterising an agent (includes login identifiers, IRC, and other chat nicknames)." ;
```

perf/data/foaf.ttl
```diff
@@ -0,0 +1,699 @@
...
+  rdfs:comment "Indicates an account held by this agent." ;
+  rdfs:domain foaf:Agent ;
+  rdfs:range foaf:OnlineAccount ;
+  rdfs:isDefinedBy foaf: .
+
+foaf:accountServiceHomepage
+  a rdf:Property, owl:ObjectProperty ;
+  ns0:term_status "testing" ;
+  rdfs:label "account service homepage" ;
+  rdfs:comment "Indicates a homepage of the service provide for this online account." ;
```
TallTed, line 644:
```suggestion
  rdfs:comment "Indicates a homepage of the service provider for this online account." ;
```

perf/data/foaf.ttl
```diff
@@ -0,0 +1,699 @@
...
+  ns0:term_status "unstable" ;
+  rdfs:label "membershipClass" ;
+  rdfs:comment "Indicates the class of individuals that are a member of a Group" ;
+  rdfs:isDefinedBy foaf: .
+
+foaf:birthday
+  a rdf:Property, owl:FunctionalProperty, owl:DatatypeProperty ;
+  ns0:term_status "unstable" ;
+  rdfs:label "birthday" ;
+  rdfs:comment "The birthday of this Agent, represented in mm-dd string form, eg. '12-31'." ;
```
TallTed, line 678:
```suggestion
  rdfs:comment "The birthday of this Agent, represented in mm-dd string form, e.g., '12-31'." ;
```

perf/data/timbl.ttl
```diff
@@ -0,0 +1,348 @@
...
+wd:Q42944 schema:name "CERN"@en.
+
+wd:Q49108 schema:name "Massachusetts Institute of Technology"@en.
+
+profi:card a foaf:PersonalProfileDocument; foaf:maker :me; foaf:primaryTopic :me.
+
+:id1618591080534
+    a solid:CurrentRole;
+    schema:description
+        "Technical strategic architecture for Inrupt. Leader of the Solid movement. Coordinate and contribute to Solid open source.";
```
TallTed, line 100: Inconsistent structure... Is this a list of nouns or verbs? Should be "Design technical & strategic architecture for Inrupt. Lead the Solid movement. Coordinate and contribute to Solid open source." or "Technical strategic architect for Inrupt. Leader of the Solid movement. Coordinator and contributor to Solid open source."

perf/data/timbl.ttl
```diff
@@ -0,0 +1,348 @@
...
+    schema:startDate "2018-04-05"^^xsd:date;
+    vcard:role "Cofounder and CTO";
+    org:member :me;
+    org:organization :id1619374832756;
+    org:role occup:7b1b5da8-573a-49bb-a38e-68725a949f4f;
+    solid:publicId wd:Q16601665 .
+:id1618785166805
+    a solid:CurrentRole;
+    schema:description
+        "The Decentralized Information Group (DIG) at MIT's Computer Science and Artificial Intelligence Laboratory (CSAIL) build all sorts of things using semantic web tech, and more recently started the Solid movement to allow people complete control of their data. ";
```
TallTed, line 110:
```suggestion
        "The Decentralized Information Group (DIG) at MIT's Computer Science and Artificial Intelligence Laboratory (CSAIL) builds all sorts of things using semantic web tech, and more recently started the Solid movement to allow people complete control of their data. ";
```

perf/data/timbl.ttl
```diff
@@ -0,0 +1,348 @@
...
+:id1619532839375 solid:publicId l:de.
+
+:id1619532904680 solid:publicId l:fr.
+
+:id1619532994614 solid:publicId l:it.
+
+:id1619628182581
+    a solid:CurrentRole;
+    schema:description
+        """Our Oxford Martin School funded Project on "Ethical Web and Data Architectures" researches Data Autonomy, Data Privacy, Algorithmic Accountability, and Data Sharing and is hiring.""";
```
TallTed, line 140:
```suggestion
        """Our Oxford Martin School funded Project on "Ethical Web and Data Architectures" researches Data Autonomy, Data Privacy, Algorithmic Accountability, and Data Sharing, and is hiring.""";
```

src/N3Reasoner.js
```diff
@@ -0,0 +1,194 @@
...
+        for (let i = 0; i < r2.premise.length; i++) {
+          const p = r2.premise[i];
+          for (const c of r1.conclusion) {
+            if (termEq(p.subject, c.subject) && termEq(p.predicate, c.predicate) && termEq(p.object, c.object)) {
+              const set = new Set();
+
+              const premise = [];
+
+              // Since these *will* be substited when we apply the rule,
+              // we need to this so that we index correctly in the subsequent section
```
TallTed, line 137:
```suggestion
              // we need do to this so that we index correctly in the subsequent section
```

src/N3Reasoner.js
```diff
@@ -0,0 +1,194 @@
...
+
+              for (let j = 0; j < r2.premise.length; j++) {
+                if (j !== i) {
+                  premise.push(getIndex(r2.premise[j], set));
+                }
+              }
+
+              // eslint-disable-next-line no-warning-comments
+              // TODO: Create new rule, with new indexing
+              //       Future, 'collapse' the next statements when the share a premise/base-premise
```
TallTed, line 150:
```suggestion
              //       Future, 'collapse' the next statements when they share a premise/base-premise
```

jeswr:
> @jeswr thanks this works. I see that the `reason` method inline updates the N3Store. Would it be possible to also support that the reason method returns a new N3Store (with the new produced quads) while keeping the old one intact.
> 
> ```
>   let newStore = N3.Reasoner(oldStore).reason([ { premise:... , conclusion: ... } ]);
> ```

It's certainly feasible, though it may incur a performance hit. I'll see if I can give this a go over the weekend (no promises though). There are, again, some API decisions to be made around this - but at the very least I would consider adding support for multiple stores as input at the same time as this would require the same additions to the logic, as that required to apply the sem-naive algorithm across the `oldStore` and the `newStore`.

perf/data/foaf.ttl
jeswr: @TallTed thanks for the comments - these are sample files (intended only for performance testing) pulled directly from https://timbl.inrupt.net/profile/card and http://xmlns.com/foaf/0.1/ respectively - if there are problems with the contents I think it would be better if you followed up with the original sources for these files.

PR #299, opened 2022-08-01
Allows readable-stream 4.0.0

Tpt:
Allows dependencies to avoid shipping both readable-stream v3 and v4.

### DIFF
--- a/package.json
+++ b/package.json
@@ -25,7 +25,7 @@
   ],
   "dependencies": {
     "queue-microtask": "^1.1.2",
-    "readable-stream": "^3.6.0"
+    "readable-stream": "^3.6.0|^4.0.0"
   },
   "devDependencies": {
     "@babel/cli": "^7.16.0",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -25,7 +25,7 @@
   ],
   "dependencies": {
     "queue-microtask": "^1.1.2",
-    "readable-stream": "^3.6.0"
+    "readable-stream": "^3.6.0|^4.0.0"
```
jeswr, line 28: Why is it necessary to keep v3?

Tpt: No, I just wanted to avoid a breaking change.

RubenVerborgh: Seems that v4 only works on Node 12 and up—we still CI on Node 10. Maybe the choice is the right solution for now indeed then.

RubenVerborgh (changes requested):
CI fails

package.json
RubenVerborgh: Dropped 10 for #300 so might as well go with 4.

RubenVerborgh:
Thanks, merged as e531951ac7f3ec085fb7faab8f1d746a0f8c2fef.

PR #301, opened 2022-08-16
Add Underscore Support in N3 Writer Prefix Matching

e2dubba:
* Update the Regex to handle an underscore for underscores in instances

### DIFF
--- a/src/N3Writer.js
+++ b/src/N3Writer.js
@@ -1,19 +1,24 @@
 // **N3Writer** writes N3 documents.
-import namespaces from './IRIs';
-import { default as N3DataFactory, Term } from './N3DataFactory';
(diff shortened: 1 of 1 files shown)


RubenVerborgh:
Hi @e2dubba! What big does this fix, and could you add a test for it?

e2dubba:
@RubenVerborgh, thanks for assigning this to me. I have it working as I expected on my local machine, but the discussion is [in this issue here](https://github.com/rdfjs/N3.js/issues/302). I'll add a couple of tests to make sure that it is producing what I expect it to.

e2dubba:
@RubenVerborgh, added a test. Please let me know if I haven't followed convention when placing this test.

RubenVerborgh (approved):

Issue #256, opened 2021-07-05
[potential bug] [RDF*] Annotated Quads not treated as Quad in store


jeswr:
This may be a misunderstanding of the intended implementation of RDF* on my part.

To my understanding, if we have a `quad` (lets call it q1) that is the subject of another `quad` then q1 should still be treated normally as a `quad`. However, this does not appear to be the case in the example below.

```ts
const data = `
@prefix :    <http://www.example.org/> .
@prefix ex:    <http://www.example.org/> .
@prefix foaf: <http://xmlns.com/foaf/0.1/> .

<<:bob foaf:age 23>> ex:certainty 0.9 .
`

const store = new Store();
const parser = new Parser({ format: 'turtle*' });

store.addQuads(parser.parse(data));


// Expected: 2, Result: 1
store.size

// Expected: 2, Result: 1
store.getQuads(null, null, null, null).length

// Expected: 1, Result: 0
store.getQuads(namedNode('http://www.example.org/bob'), null, null, null).length
```
Here I would expect that the quad with the predicate `foaf:age` and the quad with the predicate `ex:certainty` both contribute to the count, and are both searched over in the `store.getQuads` call.

RubenVerborgh:
The store doesn't do RDF-star yet as far as I recall.

rubensworks:
AFAIK this is intended behaviour. Making a statement about something does not necessarily imply that statement. (Ruben says that cows are blue =\> cows are blue)

RubenVerborgh:
I think there are two things here:
- The difference between asserted and annotated: https://w3c.github.io/rdf-star/cg-spec/editors_draft.html#turtle-star-parsing-discussion
- Whether the store supports RDF-star (I thought it didn't, might be wrong)

TallTed:
All the rest of the above notwithstanding, RDF-star (still a [draft-CG-report](https://w3c.github.io/rdf-star/), not yet even a WG Editor's Draft) isn't expected to support quoted (a/k/a reified, embedded, or inline) quads; only triples are expected to be supported in this way.

uvdsl:
I just noticed that the parser does not like RDF lists with RDF-star triples in it.

```js
const data = `
@prefix :    <http://www.example.org/> .
@prefix ex:    <http://www.example.org/> .
@prefix foaf: <http://xmlns.com/foaf/0.1/> .

:alice ex:says <<:bob foaf:age 23>> .
`
```


I get a `Error: Expected entity but got << on line 6.`.

[...]

PR #311, opened 2022-11-23
feat/rdfstar support

### LINKED ISSUE #272: Does not parse RDF lists with RDF-star triples
https://github.com/rdfjs/N3.js/issues/256#issuecomment-1024447424

Reported by @uvdsl

### LINKED ISSUE #304: Bug: `termFromId` errors on deeply nested quads
This test https://github.com/rdfjs/N3.js/blob/a8cd43609e0d83edcd1f2b64d164760e68eaebb0/test/Term-test.js#L211-L213 fails as `termFromId` does not handle nested quads. This is because this regex (https://github.com/rdfjs/N3.js/blob/e531951ac7f3ec085fb7faab8f1d746a0f8c2fef/src/N3DataFactory.js#L13) cannot handle one of the terms being a quad.

Trying to look into a fix but am struggling to get the regex for it right.

jeswr:
This PR adds full support for RDF-star with all spec tests passing.

Closes #272 
Closes #304
Related to #256 (What I consider lacking for that to be closed is specialised indexing for nested triples, and a match method that allows you to match based on patterns in nested triples).


src/N3Parser.js
```diff
@@ -614,6 +635,17 @@ export default class N3Parser {
     case ',':
       next = this._readObject;
       break;
+    case '{|':
+      if (!this._supportsRDFStar)
+        return this._error('Unexpected RDF* syntax', token);
```
TallTed, line 640: Do you really mean the deprecated `RDF*`, or should this name the current `RDF-star`?
```suggestion
        return this._error('Unexpected RDF* syntax', token);
```

src/N3Parser.js
```diff
@@ -866,6 +894,27 @@ export default class N3Parser {
     }
   }
 
+    // ### `_readRDFStarTail` reads the end of a nested RDF* triple
```
TallTed, line 897: Do you really mean the deprecated `RDF*`, or should this name the current `RDF-star`?
```suggestion
    // ### `_readRDFStarTail` reads the end of a nested RDF-star triple
```

src/N3Parser.js
```diff
@@ -835,25 +867,21 @@ export default class N3Parser {
...
-    if (token.type !== '>>') {
-      // An entity means this is a quad (only allowed if not already inside a graph)
-      if (this._supportsQuads && this._graph === null && (this._graph = this._readEntity(token)) !== undefined)
-        return this._readRDFStarTail;
-      return this._error(`Expected >> to follow "${this._object.id}"`, token);
-    }
-    return this._readRDFStarTail(token);
-  }
-
   // ### `_readRDFStarTail` reads the end of a nested RDF* triple
```
TallTed, line 870: Do you really mean the deprecated `RDF*`, or should this name the current `RDF-star`?
```suggestion
  // ### `_readRDFStarTail` reads the end of a nested RDF-star triple
```

test/N3Parser-test.js
```diff
@@ -975,14 +1126,59 @@ describe('Parser', () => {
...
+      shouldNotParse('<< <a> <b> () >> <c> <d>.',
+        'Unexpected list inside quoted triple on line 1.'
+    ));
+
+    it('should not parse non-empty list inside quoted triple object',
+      shouldNotParse('<< <a> <b> ( <f> ) >> <c> <d>.',
+        'Unexpected list inside quoted triple on line 1.'
+    ));
+
     it('should not parse nested RDF* statements that are partially closed',
```
TallTed, line 1174: Do you really mean the deprecated `RDF*`, or should this name the current `RDF-star`?
```suggestion
    it('should not parse nested RDF-star statements that are partially closed',
```
This comment applies to this whole file...

jeswr:
Thanks for the catch @TallTed - this PR essentially is changing over from implementing `RDF*` to `RDF-star`. I'd done all but a ctr+f to rename at the end :).

Also to summarise the contents of this PR:
 - Update the parser to be spec compliant
 - Update the store to handle deeply nested quads.

What is not done:
 - Changes to the writer (I'm not actually sure any are necessary).
 - Better support for matching against patterns in nested triples

TallTed:
> Update the store to handle deeply nested quads.

Note that, as of this writing, the RDF-star extension to RDF (and its serializations) is focused on triples, not quads, so this update may need reconsideration.

jeswr:
> Note that, as of this writing, the RDF-star extension to RDF (and its serializations) is focused on triples, not quads, so this update may need reconsideration.

My use of the quad terminology arises from the fact that the primitive in RDF/JS is a quad and not a triple. So here the convention is that any nested *triple* (as referred to by the spec), is emmitted as a nested RDF/JS quad where the graph term *MUST* be the default graph.

This is also enforced in the parsing changes in this PR where nested graph terms cause errors. However given that the store currently already supports things outside of the spec (e.g. literals as subjects/objects); I have made the design decision to not enforce this requirement on store operations; so nested quads in the store may contain graph terms that are not the default graph.

As I have already discussed with @rubensworks the main thing that needs to be done is to align in the RDFJS standard on a way of representing nested triples. In my view there are 3 ways of going about this:
1. Make the graph term *optional* in RDFJS quads, and do not include it in quoted triples (this would be breaking) *or* allow the graph term to be set to null (also breaking).
2. Agree that all quoted triples should be represented as RDFJS quads with the DefaultGraph set for the graph term
3. Agree that all quoted triples should be represented as RDFJS quads with the graph term the same as that of the top level quad.


To give a concrete example then if we have the following nquads statement

```nquads
<<:a :b :c>> :p :o :g
```
would, in RDFJS, become

under option 1
```ts
quad(
   quad(
      namedNode('http://example.org/a'),
      namedNode('http://example.org/b'),
      namedNode('http://example.org/c'),
      null,
   ),
  namedNode('http://example.org/p'),
  namedNode('http://example.org/o'),
  namedNode('http://example.org/g'),
)
```

under option 2 (and as is currently implemented in this PR)
```ts
quad(
   quad(
      namedNode('http://example.org/a'),
      namedNode('http://example.org/b'),
      namedNode('http://example.org/c'),
      DEFAULT_GRAPH,
   ),
  namedNode('http://example.org/p'),
  namedNode('http://example.org/o'),
  namedNode('http://example.org/g'),
)
```


under option 3
```ts
quad(
   quad(
      namedNode('http://example.org/a'),
      namedNode('http://example.org/b'),
      namedNode('http://example.org/c'),
      namedNode('http://example.org/g'),
   ),
  namedNode('http://example.org/p'),
  namedNode('http://example.org/o'),
  namedNode('http://example.org/g'),
)
```

TallTed:
> given that the store currently already supports things outside of the spec

Building to what "[a] store ... supports" rather than to what a spec requires tends to do horrible things for interop. Further, to my mind, I can't see how not building to the RDF-star spec (even as it's still a draft) cannot leave something to be desired in what you're calling "full support for RDF-star".

But perhaps I'm missing something and/or these trade-offs are OK with the community around "[this] store". I suppose time will tell.

jeswr:
> But perhaps I'm missing something and/or these trade-offs are OK with the community around "[this] store".

I think https://github.com/rdfjs/types/issues/34#issuecomment-1331408661 summarizes it nicely. That is, the data is already validated when we do the *data exchange* (in particular when we are parsing it before adding it to the store, or serializing to send it somewhere else); so there isn't a need to apply additional validation at this processing/storage stage.

A good example of where enforcing such interfaces during data processing is problematic is in https://github.com/rdfjs/N3.js/pull/296 where the reasoner is working directly with the internal index and has zero knowledge of what types of terms the id represents.

woutermont:
> [I]n this PR the convention is that any nested triple (as referred to by the spec), is emmitted as a nested RDF/JS quad where the graph term MUST be the default graph. This is also enforced in the parsing changes in this PR where nested graph terms cause errors.

@jeswr, are there specific reasons why this should be enforced, i.e. why we cannot simply ignore the graph term of quoted triples/quads without erroring?

jeswr:
> are there specific reasons why this should be enforced, i.e. why we cannot simply ignore the graph term of quoted triples/quads without erroring?

There are specific test cases (e.g. https://w3c.github.io/rdf-star/tests/turtle/syntax/#turtle-star-bad-8) forbidding this in the spec tests. We *could* opt-into allowing graph-terms and be non-compliant; but I don't see the value in maintaining such a feature.

tpluscode:
Ah, I see that the https://github.com/rdfjs/types/issues/34#issuecomment-1333452852 I just made does not align with the discussion here.

We do not want to allow graph in quoted triples after all?

woutermont:
We _do_. In this PR, as I understand it, they are just constricted to the `DEFAULT_GRAPH`.

jeswr:
> We do not want to allow graph in quoted triples after all?

When dealing with valid RDF-star statements; correct.

> We do. In this PR, as I understand it, they are just constricted to the `DEFAULT_GRAPH`.

I am just implementing option 2 in https://github.com/rdfjs/types/issues/34#issue-1467371879. That is; we set the `graph` term in the quad to `DEFAULT_GRAPH` so it is a valid RDFJS quad, but we should just ignore the graph term when reading the nested/quoted triple/quad since it is not actually part of the quoted term.

TallTed:
@woutermont 

> We _do_. In this PR, as I understand it, they are just constricted to the `DEFAULT_GRAPH`.

I suggest you change that from `We do` to `I do`, because you certainly do not speak for all participants in this thread, nor have you presented any evidence that you are speaking for _anyone_ other than yourself.

woutermont:
> @woutermont
> 
> > We _do_. In this PR, as I understand it, they are just constricted to the `DEFAULT_GRAPH`.
> 
> I suggest you change that from `We do` to `I do`, because you certainly do not speak for all participants in this thread, nor have you presented any evidence that you are speaking for _anyone_ other than yourself.

Yet @tpluscode [does speak voor everyone](https://github.com/rdfjs/N3.js/pull/311#issuecomment-1333456334)? Or do you have a bias towards me in your nitpicking, @TallTed?

TallTed:
> [@woutermont] Yet @tpluscode [does speak for everyone](https://github.com/rdfjs/N3.js/pull/311#issuecomment-1333456334)? Or do you have a bias towards me in your nitpicking, @TallTed?

No, @tpluscode doesn't speak for everyone, either.  (And if you'll peruse my activity history, even just within this PR, you should discover that my nitpicking is applied in many directions, pretty much all the time.)  However, @tpluscode _has_ generally been reflecting what others in this thread (or the related and connected other thread(s)) have also/already said, while so far as I've observed, you've been in a chorus of one.

(I do wish this were being hashed out in one place, preferably an issue under [rdf-star](https://github.com/w3c/rdf-star/issues/), perhaps even [RDF-star#275](https://github.com/w3c/rdf-star/issues/275), the results of which discussion could then be applied to the implementations in [N3.js](https://github.com/rdfjs/N3.js) and elsewhere.)

woutermont:
> @tpluscode has generally been reflecting what others in this thread (or the related and connected other thread(s)) have also/already said, while so far as I've observed, you've been in a chorus of one.

Then I'm afraid one of us is just reading the concensus in related threads wrongly 🤷‍♂️ Because I'm saying the same thing as @rubensworks and @jeswr.

src/N3DataFactory.js
```diff
@@ -188,9 +187,8 @@ export class DefaultGraph extends Term {
 // ## DefaultGraph singleton
 DEFAULTGRAPH = new DefaultGraph();
 
-
 // ### Constructs a term from the given internal string ID
-export function termFromId(id, factory) {
+export function termFromId(id, factory, nested) {
```
RubenVerborgh, line 191: @jeswr In what cases should `nested` be true? Is it the same as `rdfStar` basically?

src/N3DataFactory.js
```diff
@@ -215,21 +213,24 @@ export function termFromId(id, factory) {
...
+      return factory.namedNode(id);
+    }
   }
+  return factory.quad(
+    termFromId(id[0], factory, true),
+    termFromId(id[1], factory, true),
+    termFromId(id[2], factory, true),
+    id[3] && termFromId(id[3], factory, true)
+  );
 }
```
RubenVerborgh, line 230: Questions for the future:
- Should N3.js `Term` internally still rely on `id`?
- (For what cases) do we still need `termFromId`, or is `termToId` enough—and can it just be `.hash` on the thing itself then?

src/N3DataFactory.js
```diff
@@ -247,17 +248,17 @@ export function termToId(term) {
     term.language ? `@${term.language}` :
       (term.datatype && term.datatype.value !== xsd.string ? `^^${term.datatype.value}` : '')}`;
   case 'Quad':
-    // To identify RDF* quad components, we escape quotes by doubling them.
+    // To identify RDF-star quad components, we escape quotes by doubling them.
```
RubenVerborgh, line 251: This comment does not apply anymore then, if there's no doubling?

src/N3Lexer.js
```diff
@@ -302,12 +302,20 @@ export default class N3Lexer {
       case ')':
       case '{':
       case '}':
+        if (input.length > 1 && input[1] === '|') {
+          type = '{|', matchLength = 2;
+          break;
+        }
```
RubenVerborgh, line 308: …but that's only for `}`, right?

So we'd need
```suggestion
      case '{':
        if (input.length > 1 && input[1] === '|') {
          type = '{|', matchLength = 2;
          break;
        }
      case '{':
      case '[':
      case ']':
      case '(':
      case ')':
      case ',':
      case ';':
```

RubenVerborgh: Nah, even worse:
```suggestion
      case '{':
        if (input.length < 2) {
          inconclusive = true;
        }
        if (input[1] === '|')
          type = '{|', matchLength = 2;
          break;
        }
      case '{':
      case '[':
      case ']':
      case '(':
      case ')':
      case ',':
      case ';':
```

and have a test where the input stream breaks at `{` and the next chunk starts with `|`.

src/N3Lexer.js
```diff
@@ -302,12 +302,20 @@ export default class N3Lexer {
...
         if (!this._lineMode) {
           matchLength = 1;
           type = firstChar;
         }
         break;
-
+      case '|':
+        if (input.length > 1 && input[1] === '}') {
```
RubenVerborgh, line 315: Idem; should be inconclusive if input length smaller than 2. And we should test a split `|` and `}` (can be the same test as the split starting token).

src/N3Parser.js
```diff
@@ -614,6 +635,17 @@ export default class N3Parser {
     case ',':
       next = this._readObject;
       break;
+    case '{|':
+      if (!this._supportsRDFStar)
+        return this._error('Unexpected RDF-star syntax', token);
+
+      this._saveContext('{|', this._graph, this._subject, this._predicate, this._object);
+
+      // Note - we always use the default graph for the quoted triple component
```
RubenVerborgh, line 644:
```suggestion
      // The spec mandates the default graph for the quoted triple component
```
(I assume that is true?)

src/N3Parser.js
```diff
@@ -866,6 +894,27 @@ export default class N3Parser {
     }
   }
 
+    // ### `_readRDFStarTail` reads the end of a nested RDF-star triple
+  _readAnnotatedTail(token) {
+    if (token.type === '{|') {
+      this._saveContext('{|', this._graph, this._subject, this._predicate, this._object);
+
+      // Note - we always use the default graph for the quoted triple component
```
RubenVerborgh, line 902:
```suggestion
      // The spec mandates the default graph for the quoted triple component
```

src/N3Parser.js
```diff
@@ -866,6 +894,27 @@ export default class N3Parser {
...
+      this._subject = this._quad(this._subject, this._predicate, this._object, this.DEFAULTGRAPH);
+      this._predicate = null;
+      this._object = null;
+      return this._readPredicate;
+    }
+    else {
+      this._emit(this._subject, this._predicate, this._object, this._graph);
+    }
+
+    if (token.type !== '|}')
```
RubenVerborgh, line 912:
```suggestion
    // If the quoted triple is not finished, the next token must be a predicate
    if (token.type !== '|}')
```
(or similar)

RubenVerborgh (changes requested):
Looks great, @jeswr! Just an omission for split `{|` and `|}` tokens.

src/N3Parser.js
jeswr: No - that goes back to https://github.com/rdfjs/N3.js/pull/311#issuecomment-1330008257 and [this discussion](https://github.com/rdfjs/types/issues/34). To summarize it:
 - The rdf-star specification only has a notion of [quoted triples](https://w3c.github.io/rdf-star/cg-spec/editors_draft.html#quoted-triples), not quoted quads. Quoted triples are not asserted in any graph; they are merely a subject or object term within another triple/quad.
 - The RDF/JS spec *only* has a notion of quads; not triples.
 
 Thus we need a convention of what to do with the graph term when an RDF/JS *quad* is describing an RDF-star quoted triple. The convention in this PR is to always use the default graph on quoted triples.

src/N3Lexer.js
jeswr: Likewise we should also be testing a split in `<` and `<` in the start of quoted `<<` statements I guess?

src/N3DataFactory.js
jeswr: It is really just an *internal* parameter that is used when recursively calling `termFromId` on `quad` terms within other quads.

The only place that parameter should *ever* be used is https://github.com/rdfjs/N3.js/blob/56cc184f78761db27751aa4a8489239e0c287666/src/N3DataFactory.js#L224-L229

src/N3Parser.js
jeswr: No - see https://github.com/rdfjs/N3.js/pull/311#discussion_r1061039556

src/N3DataFactory.js
jeswr: > (For what cases) do we still need termFromId, or is termToId enough—and can it just be .hash on the thing itself then?

Any use of the `Store` will still require `termFromId` this since terms can't be recovered from a hash. Using a hash is really only an optimisation for term equality (in which case a hash for the `id` is fine).

src/N3DataFactory.js
jeswr: I've added a comment clarifying this

src/N3DataFactory.js
jeswr: Oh actually - I think I see more what you're saying is that quads could just use a set of ids/hashes in the internal representation that are then looked up.

src/N3Parser.js
```diff
@@ -641,7 +641,8 @@ export default class N3Parser {
 
       this._saveContext('{|', this._graph, this._subject, this._predicate, this._object);
 
-      // Note - we always use the default graph for the quoted triple component
+      // As a convention we use set the graph term as the Default Graph in quads representing quoted triples
```
TallTed, line 644:
```suggestion
      // As a convention, we set the graph term as the Default Graph in quads representing quoted triples
```

test/N3Parser-test.js
```diff
@@ -2564,23 +2623,109 @@ describe('Parser', () => {
...
+      splitAllWays(result, left.concat(right.substring(0, i)), right.substring(i), chunkSize);
+    }
+  }
+
+  // Return result
+  return result;
+}
+
+// Return a large number of combinations for splitting the string to test chunking - any thing with 5 characters
+// or less will test every permutation of splits possible on the string
```
TallTed, line 2645:
```suggestion
// Return a large number of combinations for splitting the string to test chunking - anything with 5 or fewer
// characters will test every permutation of splits possible on the string
```

src/N3DataFactory.js
RubenVerborgh: > Any use of the Store will still require termFromId this since terms can't be recovered from a hash.

But that could be an internal function then indeed!

src/N3DataFactory.js
RubenVerborgh: So only by itself; gotcha.

src/N3Lexer.js
RubenVerborgh: Indeed; I see that's covered there.

RubenVerborgh (commented):
Some minor things in the lexer which I'll sort out myself; then this is good to merge.

Question: did the addition of `getSplits` surface any new bugs? I've been debating with myself for ages whether or not to get this; so far, I've relied on manual additions of cases. I do like the idea and I wonder how impactful it is.

jeswr:
> But that could be an internal function then indeed! (c.f. https://github.com/rdfjs/N3.js/pull/311#discussion_r1065776281)

That's exactly what the follow up PR https://github.com/rdfjs/N3.js/pull/318 does :)

jeswr:
> Question: did the addition of getSplits surface any new bugs? I've been debating with myself for ages whether or not to get this; so far, I've relied on manual additions of cases. I do like the idea and I wonder how impactful it is.

It blew up with dozens of errors on the `{|` case you had already pointed out; and iirc it also complained at the first attempt at fixing it due to the behavior of a fall-through case. Having it pass definitely gives me confidence that the lexer is unlikely to be missing edge cases related to chunking.

jeswr:
> Some minor things in the lexer which I'll sort out myself; then this is good to merge.

Great! Let me know if there is any further work required for this PR on my end :)

jeswr:
@RubenVerborgh - when you get around to revisiting this, it is probably worth setting https://github.com/rdfjs/N3.js/blob/520054a9fb45ef48b5b58851449942493c57dace/src/N3Parser.js#L31 to true unless there is a parameter explicitly opting out. Otherwise the parser doesn't play nice with tools like `rdf-parse` which do not recognise `-star` content types.

benjaminaaron:
I've used it like this and it looks perfect in the Turtle output:

```js
const quad1 = quad(
    namedNode('http://example.org/a'),
    namedNode('http://example.org/b'),
    namedNode('http://example.org/c'),
)

const quad2 = quad(
    quad1,
    namedNode('http://example.org/d'),
    namedNode('http://example.org/e'),
)
```

Maybe it'd be worth adding a RDF-star example to the Readme? It might not be immediately clear that the way to do it is to use one quad within another one 🤔

jeswr:
@benjaminaaron - I've added an example to the writing section :)

jeswr:
@RubenVerborgh I've refactored the lexer and enabled `rdfStar` parsing by default.

I'd be inclined to merge as is and any other comments can be applied as patch releases?

PR #319, opened 2023-01-24
Support excluding graph names when serializing

### LINKED ISSUE #165: N-Quads are written if N-Triples are requested and graph is supplied

```
var N3 = require('../N3');
writer.addQuad(
);

```

Output:
```
```

LaurensRietveld:
This fixes #165

### DIFF
--- a/src/N3Writer.js
+++ b/src/N3Writer.js
@@ -35,6 +35,7 @@ export default class N3Writer {
       options = outputStream, outputStream = null;
     options = options || {};
     this._lists = options.lists;
+    this._graphs = options.graphs || 'keep';
 
     // If no output stream given, send the output as string through the end callback
     if (!outputStream) {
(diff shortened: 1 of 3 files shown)


src/N3Writer.js
```diff
@@ -84,8 +85,11 @@ export default class N3Writer {
   // ### `_writeQuad` writes the quad to the output stream
   _writeQuad(subject, predicate, object, graph, done) {
     try {
+      if (this._graphs === 'error' && !DEFAULTGRAPH.equals(graph)) {
+        done(new Error('Encountered graph name, this is forbidden.'));
```
RubenVerborgh, line 89:
```suggestion
        throw new Error('The chosen serialization settings do not support triples in a non-default graph.');
```

Have to throw here; done doesn't exit the function.

RubenVerborgh (approved):

PR #330, opened 2023-03-02
Serialize subject and predicate literals correctly

### LINKED ISSUE #329: Serializing literal subjects returns invalid results
Copy/pasting from https://github.com/CommunitySolidServer/CommunitySolidServer/issues/1582#issuecomment-1451503232

```ts
  const n3 = `"1"^^<http://www.w3.org/2001/XMLSchema#boolean> <http://example.com/p> <https://example.com/s> .`;
  const parser = new Parser({ format: 'text/n3' });
  const store = new Store(parser.parse(n3));
  const writer = new Writer({ format: 'text/turtle' });
  writer.addQuads(store.getQuads(null, null, null, null));
  writer.end((error, result): void => console.log(result));
```
Outputs
```
"1"^^http://www.w3.org/2001/XMLSchema#boolean <http://example.com/p> <https://example.com/s>.
```

Which could perhaps be excused in that it is invalid turtle so behaviour is undefined. But the same result occurs when using `text/n3` format for the writer, while this is valid N3.

joachimvh:
Closes https://github.com/rdfjs/N3.js/issues/329

This does have as disadvantage that invalid turtle will be generated if quads get added to the writer that can not be represented in turtle. If that is a problem, an alternative solution would be to have variants of the writer, similarly to how n-quads get handled, where the turtle version throws an error if it encounters a literal in the subject/predicate position and the N3 version handles it correctly. Let me know what you prefer and I can update the PR.


RubenVerborgh:
> invalid turtle will be generated if quads get added to the writer that can not be represented in turtle

I wonder if this has repercussions for #319; should we have a generic flag for "out of syntax features"?

RubenVerborgh (commented):
Perhaps we should still keep `_encodeSubject` and `_encodeObject` around, just so we can error in those functions as needed.

jeswr:
> I wonder if this has repercussions for https://github.com/rdfjs/N3.js/pull/339; should we have a generic flag for "out of syntax features"?

...perhaps, though I worry about performance degradation doing it internally; we could just supply `validTurtleQuad`, `validN3Quad` etc. functions that consumers can optionally pass their data through first to validate.

jeswr:
As an aside - are there any RDF syntaxes that don't support `a` shortand in the subject or object position? If not then the following logic could just be moved to `_encodeTerm` https://github.com/rdfjs/N3.js/blob/c7540c4c5cd4fe40447dd90a683e1ea878e75519/src/N3Writer.js#L207

joachimvh:
> As an aside - are there any RDF syntaxes that don't support `a` shortand in the subject or object position?

Turtle only allows it in the predicate position.

jeswr:
Per @jeswr, I've pushed directly to this branch (thanks to "allow edits by maintainers") to bring it up to date — @joachimvh's diagnosis and fix are untouched and remain the core of this PR; his original commit is preserved as-is.

What was added on top:
- Merged current `main` into the branch (merge, not rebase). Conflicts were limited to `_encodeQuad` (now RDF 1.2 `<<(...)>>` triple-term syntax) and the writer tests.
- Per @RubenVerborgh's earlier suggestion, `_encodeSubject` and `_encodeObject` are kept as thin wrappers around the shared literal-aware `_encodeTerm`, so they remain entry points for format-dependent errors if we want them later.
- `quadToString` predicates now also go through the literal-aware encoding.
- Additional tests: datatyped and language-tagged literal subjects, literal predicates, `text/n3` output, `quadToString` variants, and an N3 round-trip.

`npm run lint` and `npm test` pass with 100% coverage. Should be ready for @jeswr's re-review.

TallTed:
@jeswr [commented](https://github.com/rdfjs/N3.js/pull/330#issuecomment-4870197839) ... "Per @jeswr, I've pushed ... Should be ready for @jeswr's re-review."

How many @jeswr are there?

Issue #331, opened 2023-03-04
Spec testing for N3


jeswr:
N3 now has [spec-tests](https://w3c.github.io/N3/tests/) - we should use them to test the `N3.js` parser.

jeswr:
Status update: the upstream blockers are gone. rubensworks/rdf-test-suite.js#87 (tests typed as both positive Turtle and negative N3) was resolved via rubensworks/rdf-test-suite.js#89, and rdf-test-suite now officially supports the Notation3 suites (manifest-parser.ttl and manifest-extended.ttl). #333 has gone stale and conflicting in the meantime, so the remaining work is on our side: add spec-n3 scripts alongside the existing rdf11/rdf12 ones, wire them into the CI spec matrix, and triage the failures against known N3-mode parser gaps (e.g. #330). This is being implemented now — spec-n3 scripts + CI wiring PR to follow shortly. Note the suite's canonical home moved to https://w3c-cg.github.io/N3/tests (the old w3c.github.io URL 404s). Keeping this open as the tracking issue.

PR #326, opened 2023-02-27
WIP: fix n3 paths

jeswr:
This is a WIP PR to get `N3.js` to support all of the styles of paths used by EYE.

Complimentary to https://github.com/rdfjs/N3.js/pull/311/commits/2e655fba7fef05973102d45b8c352e662751cf72.

@josd are there any key test cases that you think are missing?

PR #333, opened 2023-03-09
WIP: n3 spec tests

jeswr:


jeswr:
Just discovered that a "parent" N3 manifest exists https://github.com/w3c/N3/blob/master/tests/N3Tests/manifest-parser.ttl so we should use that rather than referring to the parser and extended tests individually.

jeswr:
Also ignore the spec errors raised by paths and subject/predicate literals. They are addressed by:
 - https://github.com/rdfjs/N3.js/pull/330, and
 - https://github.com/rdfjs/N3.js/pull/326

jeswr:
As off 33be3e8 there is now a problem arising from the fact that some tests are *both* positive turtle tests and negative n3 tests which rdf-test-suite is not currently set up to support. https://github.com/rubensworks/rdf-test-suite.js/issues/87 will need to be addressed to resolve this

jeswr:
Closing this old WIP branch. The blockers it was waiting on are resolved (w3c/N3#162 merged, rubensworks/rdf-test-suite.js#87 fixed), but the spec scripts have since been restructured for the RDF 1.1/1.2 manifests, so wiring in the N3 suite is better done on a fresh branch — and the partial lexer/parser changes here (`<-`, `has`/`is`/`id`) need a proper implementation with boundary handling anyway. #331 stays open to track the actual work.

PR #335, opened 2023-03-15
fix: use BlankNode according to RDF/JS spec

### LINKED ISSUE #334: Cannot use alternative factories in n3 mode
I have never had such an issue until I tried parsing some rules with `format: 'text/n3'`. When a `factory` is used this error is thrown

> Cannot read properties of undefined (reading 'substr')
>    at N3Parser._saveContext (/node_modules/n3/lib/N3Parser.js:102:58)

I find that on that line internally the `this._graph` is accessed with properties which are not defined in the RDF/JS spec. Thus I found that no common factory implementations to work. `rdf-ext`, `@rdfjs/data-model`, nor `rdf-data-factory`

Here's an example: https://runkit.com/embed/k5l3tf4fmvpq

tpluscode:
Fixes #334 

When parsing n3 rule, the parse tries to access `.id` of a term which is not according to spec and fails when alternative factory is provided, such as `@rdfjs/data-model`. This PR's change is to the `.value` property instead

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -17,6 +17,7 @@
         "@babel/core": "^7.16.0",
         "@babel/preset-env": "^7.16.0",
         "@babel/register": "^7.16.0",
+        "@rdfjs/data-model": "^1",
         "arrayify-stream": "^1.0.0",
         "browserify": "^17.0.0",
         "chai": "^4.0.2",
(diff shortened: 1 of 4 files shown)


jeswr:
Spec tests are not part of the unit tests, have you run `npm run spec` when working locally?

tpluscode:
Yes, all pass locally.

jeswr:
@tpluscode - it looks like the problem was that the `mf:result` http://w3c.github.io/rdf-tests/turtle/IRI_spo.nt wasn't downloaded properly by the test suite - because the error

```

✖ comment_following_localName
  comment following localName
  Error: Invalid data parsing
  Input: @prefix p: <http://a.example/> .
<http://a.example/s> <http://a.example/p> p:o#comment
.


  Expected: []

  Got: [
  {
    "subject": "http://a.example/s",
    "predicate": "http://a.example/p",
    "object": "http://a.example/o",
    "graph": ""
  }
]
```

incorrectly asserts that the result should be empty when the result in the file is indeed `<http://a.example/s> <http://a.example/p> <http://a.example/o> .`. I'd suggest re-running the CI.

tpluscode:
> I'd suggest re-running the CI.

Can you do that?

jeswr:
@RubenVerborgh - can you re-run the CI on this or give me access to do so?

RubenVerborgh:
@jeswr Did both 👍

jeswr:
Closing in favour of #346

jeswr:
Aiming for this week; and the breaking changes should not require any changes in most consumers (it changes the semantics of parsing `<=` in N3 from `log:implies` to `log:isImpliedBy` in N3, enables RDF-star support by default, and doesn't expose term constructors that should have always been internal anyway).

jeswr (approved):

PR #339, opened 2023-03-24
feature/add handle quads

### LINKED ISSUE #165: N-Quads are written if N-Triples are requested and graph is supplied

jeswr:
Supercedes #319

Closes #165

### DIFF
--- a/README.md
+++ b/README.md
@@ -199,6 +199,12 @@ To write N-Triples (or N-Quads) instead, pass a `format` argument upon creation
 const writer1 = new N3.Writer({ format: 'N-Triples' });
 const writer2 = new N3.Writer({ format: 'application/trig' });
 ```
+To exclude the graph name from the serialization, pass a `graphs` argument upon creation:
+```JavaScript
+const writer3 = new N3.Writer({ graphs: 'keep' }); // Keeps the graph names (default)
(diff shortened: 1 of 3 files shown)


RubenVerborgh:
Given #330, maybe this could become:

```js
const writer3 = new N3.Writer({ syntax: 'flexible' }); // Also output quads that cannot be expressed in the chosen syntax (default)
const writer4 = new N3.Writer({ syntax: 'drop' });     // Drop components or quads that cannot be expressed in the chosen syntax
const writer5 = new N3.Writer({ graphs: 'strict' });   // Throws an error when a quad cannot be expressed in the chosen syntax
```

jeswr:
Should we add proper validation to writer at the same time (e.g. for the cases in #330)?

RubenVerborgh:
Yes, that's the direction. But we don't _need_ to have that right now; i.e., this can be the meaning of the `strict` option, and not validating everything can be a bug until fixed.

jeswr:
Per the maintainer's go-ahead, this is being finished as #638 (this branch lives on the upstream repo itself, so it could not be pushed to; @LaurensRietveld's original commit is preserved there via cherry-pick). Leaving this PR open for @jeswr to close.

PR #347, opened 2023-03-28
quad spec usage

jeswr:
- fix: use BlankNode according to RDF/JS spec
- chore: add turtle tests for external data factory


jeswr:
Supercedes #335 (had to create a new PR to trick github into running the required coveralls action)

PR #354, opened 2023-05-30
Update README.md

mimfa:
There I added some comments

### DIFF
--- a/README.md
+++ b/README.md
@@ -75,7 +75,7 @@ console.log(myQuad.object.language);       // en
 ```
 
 In the rest of this document, we will treat “triples” and “quads” equally:
-we assume that a quad is simply a triple in a named or default graph ("_").
+we assume that a quad is simply a triple in a named or default graph.
 
 ## Parsing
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -177,16 +177,16 @@ A dedicated `prefix` event signals every prefix with `prefix` and `term` argumen
 Write quads through `addQuad`.
 
 ```JavaScript
-const writer = new N3.Writer({ prefixes: { c: 'http://example.org/cartoons#' } });
+const writer = new N3.Writer({ prefixes: { c: 'http://example.org/cartoons#' } }); // Create a writer with a special context as Graph
```
jeswr, line 180:
```suggestion
const writer = new N3.Writer({ prefixes: { c: 'http://example.org/cartoons#' } }); // Create a writer which uses `c` as a prefix for the namespace http://example.org/cartoons#
```

jeswr (changes requested):

README.md
TallTed:
```suggestion
const writer = new N3.Writer({ prefixes: { c: 'http://example.org/cartoons#' } }); // Create a writer which uses `c` as a prefix for the namespace `http://example.org/cartoons#`
```

jeswr (approved):

PR #341, opened 2023-03-24
versions/2.0.0

jeswr:
WIP For version 2 of N3.js

TODO:
 - Make lists and similar entites work in arbitrary path positions (in N3)
 - Predicate literals?
 - Error on subject lists in turtle?

PR #358, opened 2023-06-22
Add support for the annotated triples syntax

rubensworks:
The RDF-star spec introduces syntatical suggar for annotating triples: https://w3c.github.io/rdf-star/cg-spec/editors_draft.html#annotation-syntax

### DIFF
--- a/src/N3Lexer.js
+++ b/src/N3Lexer.js
@@ -300,13 +300,37 @@ export default class N3Lexer {
       case ']':
       case '(':
       case ')':
-      case '{':
       case '}':
         if (!this._lineMode) {
           matchLength = 1;
           type = firstChar;
+          break;
+        }
+      case '{':
+        if (!this._lineMode) {
+          // We need at least 2 tokens lookahead to distinguish "{|" and "{ "
+          if (input.length < 2)
+            break;
+
+          // Try to find a quoted triple annotation start
+          if (input.length > 1 && input[1] === '|') {
+            type = '{|', matchLength = 2;
+            break;
+          }
+
+          matchLength = 1;
+          type = firstChar;
         }
         break;
+      case '|':
+        // We need at least 2 tokens lookahead to distinguish "|}" and "|"
+        if (input.length < 2)
+          break;
+        // Try to find a quoted triple annotation end
+        if (input[0] === '|' && input.length > 1 && input[1] === '}') {
+          type = '|}', matchLength = 2;
+          break;
+        }
 
       default:
         inconclusive = true;
--- a/src/N3Lexer.js
+++ b/src/N3Lexer.js
@@ -323,7 +323,7 @@ export default class N3Lexer {
         }
         break;
       case '|':
-        // We need at least 2 tokens lookahead to distinguish "|}" and "|"
+        // We need 2 tokens lookahead to parse "|}"
         if (input.length < 2)
           break;
         // Try to find a quoted triple annotation end
--- a/src/N3Lexer.js
+++ b/src/N3Lexer.js
@@ -324,13 +324,10 @@ export default class N3Lexer {
         break;
       case '|':
         // We need 2 tokens lookahead to parse "|}"
-        if (input.length < 2)
-          break;
         // Try to find a quoted triple annotation end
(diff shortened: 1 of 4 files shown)


jeswr:
Note that support for this is also added as part of https://github.com/rdfjs/N3.js/pull/341; however I am in favor of releasing this as PR as a new minor version - and I will aim to get the next mver out soon.

src/N3Lexer.js
```diff
@@ -300,13 +300,37 @@ export default class N3Lexer {
...
+            break;
+
+          // Try to find a quoted triple annotation start
+          if (input.length > 1 && input[1] === '|') {
+            type = '{|', matchLength = 2;
+            break;
+          }
+
+          matchLength = 1;
+          type = firstChar;
```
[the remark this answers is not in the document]
joachimvh: I highly disagree that any situation makes it a good idea. I would actually go for the following here:
```ts
type = input[1] === '|' ? '{|' : '{'}
matchLength = type.length;
```
Although I guess checking the length of type degrades performance by a microsecond. If that matters I would still go for an if/else block with 2 lines each without the comma.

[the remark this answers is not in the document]
rubensworks: To be honest, I'm also not a big fan of the comma operator (as well as if-else statements without curly brackets), just using them here for consistency :-)

src/N3Lexer.js
```diff
@@ -300,12 +300,33 @@ export default class N3Lexer {
...
+        if (!this._lineMode) {
+          // We need at least 2 tokens lookahead to distinguish "{|" and "{ "
+          if (input.length < 2)
+            break;
+
+          // Try to find a quoted triple annotation start
+          if (input[1] === '|') {
+            type = '{|', matchLength = 2;
+            break;
+          }
```
jeswr, line 319: Having this inside the `!this._lineMode` condition implies that annotation syntax is not supported in the n-quads flavour of RDF-star.

Is this correct?

Note this `{` case is currently implemented as follows in the staged v2 branch

```ts
        // We can properly evaluate this case if we are
        // not in rdfStar mode or we can look ahead to
        // see if there is a pipe following the {
        const long = input.length !== 1;
        if (long && input[1] === '|')
          matchLength = 2, type = '{|';
        else if (!this._lineMode && (long || !this._supportsRDFStar))
          matchLength = 1, type = firstChar;
        break;
```

rubensworks: AFAICS, N-Triples and N-Quads do not support annotated triples: https://w3c.github.io/rdf-star/cg-spec/editors_draft.html#n-triples-star

This makes sense, given that annotated triples are just syntactic sugar, and N-Triples and N-Quads have not provided the previously-existing syntactical shortcuts that Turtle and TriG had so far.

jeswr: Makes sense; and the [n-quads test suite](https://w3c.github.io/rdf-star/tests/nt/syntax/) doesn't test for it.

I've opened up https://github.com/w3c/rdf-star/issues/279 requesting some negative tests for this as well.

src/N3Lexer.js
```diff
@@ -300,12 +300,33 @@ export default class N3Lexer {
...
+
+          matchLength = 1;
+          type = firstChar;
+        }
```
jeswr, line 323:
```suggestion
        // We need at least 2 tokens lookahead to distinguish "{|" and "{ "
        if (!this._lineMode && input.length >= 2) {
          // Try to find a quoted triple annotation start
          if (input[1] === '|')
            type = '{|', matchLength = 2;
          else
            type = firstChar, matchLength = 1;
        }
```

src/N3Lexer.js
```diff
@@ -300,13 +300,28 @@ export default class N3Lexer {
       case ']':
       case '(':
       case ')':
-      case '{':
       case '}':
         if (!this._lineMode) {
           matchLength = 1;
           type = firstChar;
+          break;
+        }
```
jeswr, line 308:
```suggestion
        }
        break;
```

This will just end up hitting the break on the next case anyways.

jeswr (approved):

jeswr:
Released as v1.17.0 - thanks @rubensworks !

PR #366, opened 2023-10-18
chore: migrate to jest

jeswr:
Largely automated transformation using [jest-codemods](https://jestjs.io/docs/migration-guide) - originally did this to get better test diffs using jest-rdf to analyze https://github.com/rdfjs/N3.js/pull/364 but it turns out the spec tests are driven differently.


RubenVerborgh (approved):
Great work, thank you!

PR #369, opened 2023-10-20
fix: indexing quoted triples

jeswr:
This work was originally in https://github.com/rdfjs/N3.js/pull/341, but have pulled into a separate PR as this can be put in a patch release.

There are no performance degredations against `N3Store-perf.js`

### DIFF
--- a/test/N3Store-test.js
+++ b/test/N3Store-test.js
@@ -1,11 +1,13 @@
 import {
   Store,
+  termFromId, termToId,
+} from '../src/';
+import {
   NamedNode,
   Literal,
   DefaultGraph,
   Quad,
-  termFromId, termToId,
-} from '../src/';
+} from '../src/N3DataFactory';
 import namespaces from '../src/IRIs';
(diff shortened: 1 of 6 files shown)


test/N3Store-test.js
```diff
@@ -299,43 +301,119 @@ describe('Store', () => {
...
+        stream.on('end', () => {
+          expect(s2.size).toEqual(2);
+          done();
+        });
+      }
+    );
+  });
+
+  // These tests should probably be broken in the future; they are here to serve to use that we should to a mver bump
+  // at such a time
```
TallTed, line 376: The portion following the semicolon needs rewriting. At present, it is indecipherable.

TallTed (changes requested):

test/N3Store-test.js
jeswr, line 376:
```suggestion
```

RubenVerborgh (approved):
Okay that's quite cool 😎 Looks good to me!

PR #370, opened 2023-10-27
chore: semantic release

jeswr:
Adds a semantic release mechanism (releases are made based off semantic `fix`/`feat`/`BREAKING CHANGE` commit messages to main).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -70,3 +70,25 @@ jobs:
         with:
           github_token: ${{ secrets.GITHUB_TOKEN }}
           publish_dir: .
+
+  release:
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -42,6 +42,7 @@
     "pre-commit": "^1.2.2",
     "rdf-isomorphic": "^1.3.1",
     "rdf-test-suite": "^1.25.0",
+    "semantic-release": "^21.1.2",
```
RubenVerborgh, line 45: We don't need this though, right?

RubenVerborgh (approved):
Scary, but yes let's do this 👍

PR #374, opened 2023-12-06
Add support for empty graphs in the subject and object positions

Dexagod:
This PR contains a fix for parsing empty graphs in subject and object positions when parsing empty graphs.

I added some test cases that check for empty graphs in the subject and object position, both in the default graph, and contained in a subgraph.

Note that this does NOT work for empty graphs in the predicate position.
I have added tests for these as well, but they are commented out at the moment.

### DIFF
--- a/src/N3Parser.js
+++ b/src/N3Parser.js
@@ -663,8 +663,23 @@ export default class N3Parser {
     const graph = this._graph;
 
     // Store the last quad of the formula
-    if (this._subject !== null)
-      this._emit(this._subject, this._predicate, this._object, graph);
+
+    if (this._subject !== null) {
+      // Catch the empty graph being closed when parsing N3.
+      // In this case, we emit the empty graph as the value "true"^^xsd:boolean
+      if (!this._predicate && !this._object) {
+        const outerGraph = this._contextStack[this._contextStack.length - 1].graph;
+        this._emit(
+          graph,
+          this._namedNode('http://www.w3.org/1999/02/22-rdf-syntax-ns#value'),
+          this._literal('true', this._namedNode('http://www.w3.org/2001/XMLSchema#boolean')),
+          outerGraph
+        ); // Restore the parent context containing this formula
+      }
+      else {
+        this._emit(this._subject, this._predicate, this._object, graph);
+      }
+    }
 
     // Restore the parent context containing this formula
     this._restoreContext('formula', token);
--- a/src/N3Parser.js
+++ b/src/N3Parser.js
@@ -662,19 +662,17 @@ export default class N3Parser {
 
     const graph = this._graph;
 
+    let closingEmptyGraph = false;
+
     // Store the last quad of the formula
 
     if (this._subject !== null) {
       // Catch the empty graph being closed when parsing N3.
       // In this case, we emit the empty graph as the value "true"^^xsd:boolean
-      if (!this._predicate && !this._object) {
-        const outerGraph = this._contextStack[this._contextStack.length - 1].graph;
-        this._emit(
-          graph,
-          this._namedNode('http://www.w3.org/1999/02/22-rdf-syntax-ns#value'),
-          this._literal('true', this._namedNode('http://www.w3.org/2001/XMLSchema#boolean')),
-          outerGraph
-        ); // Restore the parent context containing this formula
+      // The predicate value is set to 'null' in the _readSubject function,
+      // after which the graph is closed directly when processing an empty graph.
+      if (!this._predicate) {
+        closingEmptyGraph = true;
       }
       else {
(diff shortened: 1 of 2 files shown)


RubenVerborgh:
@Dexagod Are empty graphs supported by both N3 (seems yes) and RDF-star (citation needed)?
If not, we will need a check on content type.

RubenVerborgh (changes requested):
It's definitely not `true` in subject position for RDF-star, so we will need a format check.

RubenVerborgh:
(I might be overcautious here, and maybe this code doesn't hit with RDF-star; just want to be sure.)

src/N3Parser.js
```diff
@@ -663,8 +663,23 @@ export default class N3Parser {
...
-      this._emit(this._subject, this._predicate, this._object, graph);
+
+    if (this._subject !== null) {
+      // Catch the empty graph being closed when parsing N3.
+      // In this case, we emit the empty graph as the value "true"^^xsd:boolean
+      if (!this._predicate && !this._object) {
+        const outerGraph = this._contextStack[this._contextStack.length - 1].graph;
+        this._emit(
+          graph,
+          this._namedNode('http://www.w3.org/1999/02/22-rdf-syntax-ns#value'),
```
RubenVerborgh, line 674: Probably want to use a constant for this.

src/N3Parser.js
```diff
@@ -663,8 +663,23 @@ export default class N3Parser {
...
+          this._literal('true', this._namedNode('http://www.w3.org/2001/XMLSchema#boolean')),
```
RubenVerborgh, line 675: Probably want to use a constant for this.

src/N3Parser.js
```diff
@@ -663,8 +663,23 @@ export default class N3Parser {
     const graph = this._graph;
 
     // Store the last quad of the formula
-    if (this._subject !== null)
-      this._emit(this._subject, this._predicate, this._object, graph);
+
+    if (this._subject !== null) {
+      // Catch the empty graph being closed when parsing N3.
+      // In this case, we emit the empty graph as the value "true"^^xsd:boolean
+      if (!this._predicate && !this._object) {
```
RubenVerborgh, line 670: Is it sufficient to check that there is no predicate?

jeswr:
@Dexagod indeed we will only want to do this in N3 mode and error otherwise.

@RubenVerborgh this PR is not about RDF-Star; it is specifically for N3 content types in order to align with the decisions made in https://github.com/w3c/N3/issues/185 & https://github.com/rdfjs/types/issues/37 in the next major release.

RubenVerborgh:
Ack about it being for N3; just want to make sure it doesn't fire in the wrong moments. But perhaps `_readFormulaTail` is only accessed in N3 mode in any case.

Dexagod:
Yep, synced on it with Jos and I indeed was mistaken. I'm setting it straight atm

Dexagod:
Empty graphs are now correctly represented as "true"^^xsd:boolean.

I checked and the `_readFormulaTail` function is only called in N3 mode.

Should I check if RDF-star is enabled and throw an error in case it is?

src/N3Parser.js
Dexagod: This was an error on my part.
I represented the empty graph as a blank node with an `rdf:value` of `"true"^^xsd:boolean`,
whereas I was corrected that it should be directly replaced with `"true"^^xsd:boolean`.

jeswr:
> Should I check if RDF-star is enabled and throw an error in case it is?

No, I do not think we should error when N3 and star are both enabled. There is just no particular behavior we have to comply with as there is no N3-star spec.

src/N3Parser.js
RubenVerborgh: Okay. so let's do `if (this._predicate === null)` indeed.

jeswr:
Thanks @Dexagod — the direction here was right, and this has now been picked up: #618 reimplements it against current main. Two things made landing this branch as-is impossible: it targets `versions/2.0.0`, which was superseded (v2 shipped from `main` via the RDF 1.2 rewrite, so the patch no longer applies), and the pre-`_restoreContext` `_subject !== null` heuristic only fired for the first statement of a document (`<p> <q> <r>. {} <b> <c>.` still produced a blank-node subject).

#618 also splits the fix per review: the stale-subject/garbage-quad bug is fixed unconditionally, while the `{}` → `"true"^^xsd:boolean` mapping — following the w3c-cg/N3#185 direction and the N3 spec tests from w3c-cg/N3#202, two of which are ported as unit tests — is opt-in until the next major (tracked in #632). Credit for the original approach is kept there. Closing this as superseded; please follow along on #618.

Issue #386, opened 2024-06-10
[Support or Bug] Uncaught TypeError: process.nextTick is not a function


white-gecko:
When I use the n3 `StreamParser` in a vue application with vite in the browser I get the error:

```
Uncaught TypeError: process.nextTick is not a function
```

I think this must be a quite common setup, so I wonder wham am I doing wrong?

I have put my mwe in a repository: https://github.com/white-gecko/2024-06-10-n3-issue-mwe

If you want to have more information please ask.

Thank you very much.

jeswr:
Looking at your error log, the error seems to be occurring in the `readable-stream` dependency in https://github.com/nodejs/readable-stream/blob/main/lib/internal/streams/readable.js - would you be able to create a minimal repro making use of that package directly without N3.js and open an issue upstream.

RubenVerborgh:
Suspected compilation pipeline misconfiguration; the `process.nextTick` bit is Node.js code that should never end up in the browser.

bergos:
I'm guilty of adding the `process` Node.js polyfill to the `readable-stream` package. It was required because most bundlers removed the feature to inject the polyfills automatically.

I've used `readable-stream` with different bundlers, including `vite`, without any problems. But I haven't used it in the context of a Vue application. Maybe it's overwritten by Vue. I also noticed that you load the Node.js polyfills. `n3` doesn't require it. If you added it for `streamify-string`, then you can get rid of it with `Readable.from([dataString])`.

bergos:
I guess the polyfills are older. That's why there is no `Readable.from()` method. You must add `readable-stream` as a dependency in the latest version and use it in the `import` statement. Modern bundlers don't give you any packages out-of-the-box. All imports should point to your source or packages that are part of your dependencies.

When using `Readable.from()`, don't forget that it must be called with an array/iterable. Just wrap the string as a single element into an array like this: `[dataString]`.

white-gecko:
@bergos Thank you for your help. Still I don't get it running.

When I replace `streamify-string` by `readable-string` cf. https://github.com/white-gecko/2024-06-10-n3-issue-mwe/pull/1/files I get the `nextTick` error.

```
Uncaught TypeError: process.nextTick is not a function
    resume readable.js:903
    resume readable.js:895
    on readable.js:821
    pipe readable.js:702
    tryStore App.vue:24
    mounted App.vue:18
    VueJS 8
    <anonymous> main.js:4
readable.js:903:12
```

[...]

PR #387, opened 2024-06-22
fix: added buffer dependency

bergos:
I found one line in the code where `Buffer` is used. `Buffer` is only in Node.js environments a global object and not available in the browser. Most of the time, this is probably not a problem because `readable-stream` imports the package, but in the wrong order, this may cause problems. I stumbled over this after reading the [discussion](https://github.com/rdfjs/N3.js/issues/386#issuecomment-2163881350) of recently opened issue.

This PR adds `buffer` as a dependency and adds an import in the file where `Buffer.concat()` is used. In browser environments, nothing changes. It only makes the usage of `buffer` explicit. In Node.js or other compatible environments, the implementation of `Buffer.concat()` will be a different one, but the package is from a trusted author, it is performance optimized, and it's already used by the nested dependency `readable-stream`.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -198,7 +198,7 @@
     'import/no-amd': 2,
     'import/no-commonjs': 2,
     'import/no-import-module-exports': 2,
-    'import/no-nodejs-modules': 2,
+    'import/no-nodejs-modules': 0,
     'import/unambiguous': 2,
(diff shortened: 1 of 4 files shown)


RubenVerborgh (approved):
Great work, thanks!

PR #288, opened 2022-04-16
WIP: add full dataset functionality

jeswr:
Draft PR to add full functionality of https://github.com/rdfjs/types/blob/master/dataset.d.ts.

TODO:
 - Implement `toCanonical` and `toString`
 -  Add tests
 - Reflect these methods in `DatasetCoreAndReadableStream`
 - Documentation

PR #392, opened 2024-07-28
feat: Implement RDF/JS dataset specification

### LINKED ISSUE #390: Support for set operations on graphs/datasets in the N3Store
Hi, I was wondering if it's within the scope of N3.js to support set operations on graphs or datasets within the N3Store?

If I am able to contribute an implementation, would you accept it? Or would an extension like this fall outside the scope of N3.js?

### LINKED ISSUE #391: Why does `#every` return false on the empty store
As shown in the test here `#every` returns `false` when called on an empty store. There is a test enforcing this behavior at https://github.com/rdfjs/N3.js/blob/0ba9ec2e0e7d812f9e86985de3966583d228e189/test/N3Store-test.js#L92 This does not align with set semantics where `#every` would align with `forAll` and thus be vacuously `true` on the empty set.

It also goes against the RDF/JS Dataset definition of for `#every` https://rdf.js.org/dataset-spec/#dom-dataset-every; and so in order to complete https://github.com/rdfjs/N3.js/pull/288 we would need to alter this behavior and decide whether this change is (1) breaking or (2) a bug fix - despite the fact that there is a test enforcing the buggy behavior. @RubenVerborgh what do you think?

jeswr:
Draft PR to add full functionality of https://github.com/rdfjs/types/blob/master/dataset.d.ts. Observe that we alter the behavior of `#empty` on the empty set to now return `true` instead of `false`, as described in https://github.com/rdfjs/N3.js/issues/391 - so it may be that this needs to be an mver bump.

Supercedes https://github.com/rdfjs/N3.js/pull/288, pushing to a branch on the N3.js repo so CI will run.

Closes #390
Closes #391

### DIFF
--- a/README.md
+++ b/README.md
@@ -305,7 +305,7 @@ for (const quad of store.match(namedNode('http://ex.org/Mickey'), null, null))
 ```
 
 ### [`Dataset` Interface](https://rdf.js.org/dataset-spec/#dataset-interface)
-This store adheres to the `Datase` interface which exposes the following properties
+This store adheres to the `Dataset` interface which exposes the following properties
 
 Attributes:
  - `size` — A non-negative integer that specifies the number of quads in the set.
--- a/README.md
+++ b/README.md
@@ -304,8 +304,8 @@ for (const quad of store.match(namedNode('http://ex.org/Mickey'), null, null))
   console.log(quad);
 ```
 
-### [`DatasetCore` Interface](https://rdf.js.org/dataset-spec/#datasetcore-interface)
-This store adheres to the `DatasetCore` interface which exposes the following properties
+### [`Dataset` Interface](https://rdf.js.org/dataset-spec/#dataset-interface)
+This store adheres to the `Datase` interface which exposes the following properties
 
 Attributes:
  - `size` — A non-negative integer that specifies the number of quads in the set.
@@ -318,7 +318,7 @@ Methods:
  - `[Symbol.iterator]` — Implements the iterator protocol to allow iteration over all `quads` in the dataset as in the example above.
 
 ### Addition and deletion of quads
-The store provides the following manipulation methods in addition to implementing the standard [`DatasetCore` Interface](https://rdf.js.org/dataset-spec/#datasetcore-interface)
+The store provides the following manipulation methods in addition to implementing the standard [`Dataset` Interface](https://rdf.js.org/dataset-spec/#dataset-interface)
 ([documentation](http://rdfjs.github.io/N3.js/docs/N3Store.html)):
 - `addQuad` to insert one quad
 - `addQuads` to insert an array of quads
--- a/README.md
+++ b/README.md
@@ -318,7 +318,7 @@ Methods:
  - `[Symbol.iterator]` — Implements the iterator protocol to allow iteration over all `quads` in the dataset as in the example above.
 
 ### Addition and deletion of quads
-The store provides the following manipulation methods in addition to implementing the standard [`Dataset` Interface](https://rdf.js.org/dataset-spec/#dataset-interface)
+The store implements the following manipulation methods in addition to the standard [`Dataset` Interface](https://rdf.js.org/dataset-spec/#dataset-interface)
 ([documentation](http://rdfjs.github.io/N3.js/docs/N3Store.html)):
 - `addQuad` to insert one quad
 - `addQuads` to insert an array of quads
--- a/src/N3Store.js
+++ b/src/N3Store.js
@@ -871,13 +871,10 @@ export default class N3Store {
    * This method is aligned with `Array.prototype.reduce()` in ECMAScript-262.
    */
   reduce(callback, initialValue) {
-    let accumulator = initialValue;
-    for (const quad of this) {
-      if (accumulator === undefined)
-        accumulator = quad;
-      else
-        accumulator = callback(accumulator, quad, this);
-    }
+    const iter = this.readQuads();
+    let accumulator = initialValue === undefined ? iter.next().value : initialValue;
+    for (const quad of iter)
+      accumulator = callback(accumulator, quad, this);
     return accumulator;
   }
 
--- a/src/N3Store.js
+++ b/src/N3Store.js
@@ -3,6 +3,7 @@ import { Readable } from 'readable-stream';
 import { default as N3DataFactory, termToId, termFromId } from './N3DataFactory';
 import namespaces from './IRIs';
 import { isDefaultGraph } from './N3Util';
+import N3Writer from './N3Writer';
 
 // ## Constructor
 export default class N3Store {
@@ -799,7 +800,11 @@ export default class N3Store {
    * Blank Nodes will be normalized.
    */
   contains(other) {
-    return other.every(quad => this.has(quad));
+    console.log('containes called on')
+    return other.every(quad => {
+      console.log('every called on', quad);
+      return this.has(quad)
+    });
   }
 
   /**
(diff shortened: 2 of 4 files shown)


src/N3Store.js
```diff
@@ -775,6 +771,158 @@ export default class N3Store {
...
+      this.removeQuad(quad);
+    return this;
+  }
+
+  /**
+   * Returns a new dataset that contains all quads from the current dataset, not included in the given dataset.
+   */
+  difference(other) {
+    return this.filter(quad => !other.has(quad));
+  }
```
jeswr, line 823: There is lots of room for optimisation here; in particular, whenever `other` is also an `N3Store` and they share an index we can just merge / diff / compare the indices for set operations.

Would be worth opening up an issue to do this once this PR is merged.

README.md
```diff
@@ -318,7 +318,7 @@ Methods:
  - `[Symbol.iterator]` — Implements the iterator protocol to allow iteration over all `quads` in the dataset as in the example above.
 
 ### Addition and deletion of quads
-The store provides the following manipulation methods in addition to implementing the standard [`DatasetCore` Interface](https://rdf.js.org/dataset-spec/#datasetcore-interface)
+The store provides the following manipulation methods in addition to implementing the standard [`Dataset` Interface](https://rdf.js.org/dataset-spec/#dataset-interface)
```
TallTed, line 321:
```suggestion
The store implements the following manipulation methods in addition to the standard [`Dataset` Interface](https://rdf.js.org/dataset-spec/#dataset-interface)
```

src/N3Store.js
```diff
@@ -775,6 +771,158 @@ export default class N3Store {
     return lists;
   }
 
+  /**
+   * Returns `true` if the current instance is a superset of the given dataset; differently put: if the given dataset
+   * is a subset of, is contained in the current dataset.
```
TallTed, line 776:
```suggestion
   * Returns `true` if the current dataset is a superset of the given dataset; in other words, returns `true` if
   * the given dataset is a subset of, i.e., is contained within, the current dataset.
```

src/N3Store.js
```diff
@@ -775,6 +771,158 @@ export default class N3Store {
...
+      for (const quad of quads)
+        this.add(quad);
+    }
+    return this;
+  }
+
+
+  /**
+   * Returns `true` if the current instance is a superset of the given dataset; differently put: if the given dataset
+   * is a subset of, is contained in the current dataset.
```
TallTed, line 793:
```suggestion
   * Returns `true` if the current dataset is a superset of the given dataset; in other words, returns `true` if
   * the given dataset is a subset of, i.e., is contained within, the current dataset.
```

src/N3Store.js
```diff
@@ -775,6 +771,158 @@ export default class N3Store {
...
+   *
+   * Blank Nodes will be normalized.
+   */
+  contains(other) {
+    return other.every(quad => this.has(quad));
+  }
+
+  /**
+   * This method removes the quads in the current instance that match the given arguments.
```
TallTed, line 802:
```suggestion
   * This method removes the quads in the current dataset that match the given arguments.
```

src/N3Store.js
```diff
@@ -775,6 +771,158 @@ export default class N3Store {
...
+   *
+   * The logic described in {@link https://rdf.js.org/dataset-spec/#quad-matching|Quad Matching} is applied for each
+   * quad in this dataset to select the quads which will be deleted.
```
TallTed, line 805:
```suggestion
   * quad in this dataset, to select the quads which will be deleted.
```

src/N3Store.js
```diff
@@ -775,6 +771,158 @@ export default class N3Store {
...
+   * @param graph     The optional exact graph to match.
+   */
+  deleteMatches(subject, predicate, object, graph) {
+    for (const quad of this.match(subject, predicate, object, graph))
+      this.removeQuad(quad);
+    return this;
+  }
+
+  /**
+   * Returns a new dataset that contains all quads from the current dataset, not included in the given dataset.
```
TallTed, line 819:
```suggestion
   * Returns a new dataset that contains all quads from the current dataset that are not included in the given dataset.
```

src/N3Store.js
```diff
@@ -775,6 +771,158 @@ export default class N3Store {
...
+   */
+  difference(other) {
+    return this.filter(quad => !other.has(quad));
+  }
+
+  /**
+   * Returns true if the current instance contains the same graph structure as the given dataset.
```
TallTed, line 826:
```suggestion
   * Returns true if the current dataset contains the same graph structure as the given dataset.
```

src/N3Store.js
```diff
@@ -775,6 +771,158 @@ export default class N3Store {
...
+  filter(iteratee) {
+    const store = new N3Store();
+    for (const quad of this)
+      if (iteratee(quad, this))
+        store.add(quad);
+    return store;
+  }
+
+  /**
+   * Returns a new dataset containing alls quads from the current dataset that are also included in the given dataset.
```
TallTed, line 848:
```suggestion
   * Returns a new dataset containing all quads from the current dataset that are also included in the given dataset.
```

src/N3Store.js
```diff
@@ -775,6 +771,158 @@ export default class N3Store {
...
+    const store = new N3Store();
+    for (const quad of this)
+      store.add(iteratee(quad, this));
+    return store;
+  }
+
+  /**
+   * This method calls the `iteratee` on each `quad` of the `Dataset`. The first time the `iteratee` is called, the
+   * `accumulator` value is the `initialValue` or, if not given, equals to the first quad of the `Dataset`. The return
+   * value of the `iteratee` is used as `accumulator` value for the next calls.
```
TallTed, line 867:
```suggestion
   * This method calls the `iteratee` method on each `quad` of the `Dataset`. The first time the `iteratee` method
   * is called, the `accumulator` value is the `initialValue`, or, if not given, equals the first quad of the `Dataset`.
   * The return value of each call to the `iteratee` method is used as the `accumulator` value for the next call.
```

src/N3Store.js
```diff
@@ -775,6 +771,158 @@ export default class N3Store {
...
+  reduce(callback, initialValue) {
+    const iter = this.readQuads();
+    let accumulator = initialValue === undefined ? iter.next().value : initialValue;
+    for (const quad of iter)
+      accumulator = callback(accumulator, quad, this);
+    return accumulator;
+  }
+
+  /**
+   * Returns the set of quads within the dataset as a host language native sequence, for example an `Array` in
```
TallTed, line 882:
```suggestion
   * Returns the set of quads within the dataset as a host-language-native sequence, for example an `Array` in
```

src/N3Store.js
```diff
@@ -775,6 +771,158 @@ export default class N3Store {
...
+   * ECMAScript-262.
+   *
+   * Since a `Dataset` is an unordered set, the order of the quads within the returned sequence is arbitrary.
+   */
+  toArray() {
+    return this.getQuads();
+  }
+
+  /**
+   * Returns an N-Quads string representation of the dataset, preprocessed with
```
TallTed, line 892:
```suggestion
   * Returns an N-Quads string representation of the dataset, preprocessed with the
```

TallTed (changes requested):
Minor, editorial, human-facing

RubenVerborgh (approved):
Yes, this is great!

src/N3Store.js
jeswr: @TallTed all phrasing for JSDoc in this PR is taken directly from https://rdf.js.org/dataset-spec/#dataset-interface. You may wish to reflect these changes over there.

PR #394, opened 2024-07-28
feat: shared entity index

jeswr:
Implements a shared entity index to reduce memory consumption across multiples stores.

### DIFF
--- a/README.md
+++ b/README.md
@@ -304,6 +304,13 @@ for (const quad of store.match(namedNode('http://ex.org/Mickey'), null, null))
   console.log(quad);
 ```
 
+If you are using multiple stores then you can reduce memory consumption by allowing them to share an entity index:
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -304,6 +304,13 @@ for (const quad of store.match(namedNode('http://ex.org/Mickey'), null, null))
   console.log(quad);
 ```
 
+If you are using multiple stores then you can reduce memory consumption by allowing them to share an entity index:
```
TallTed, line 307:
```suggestion
If you are using multiple stores, you can reduce memory consumption by allowing them to share an entity index:
```

TallTed (changes requested):
Minor, editorial, human-facing

RubenVerborgh (approved):
Nice 🥳

PR #399, opened 2024-07-29
perf: fix backpressure on readables

jeswr:
Fixes backpressure issues on `Readable`s; currently when streams are used, the entire dataset is put into the buffer before the stream starts emitting elements.


RubenVerborgh (approved):
Oh wow, yeah that wasn't good!

PR #401, opened 2024-07-29
perf: match over index

jeswr:


README.md
```diff
@@ -304,6 +304,13 @@ for (const quad of store.match(namedNode('http://ex.org/Mickey'), null, null))
   console.log(quad);
 ```
 
+If you are using multiple stores then you can reduce memory consumption by allowing them to share an entity index:
```
TallTed, line 307:
```suggestion
If you are using multiple stores, you can reduce memory consumption by allowing them to share an entity index:
```

TallTed (changes requested):
Editorial, minor, human-facing punctuation

jeswr:
Closing this one out: the goal here (performing match() over the internal indices rather than materialising quads) landed via #443 and was further improved in #457 and #458, on top of the shared entity index from #394. The remaining diff on this branch is a stale WIP snapshot that conflicts with main and adds nothing beyond what has already merged.

PR #417, opened 2024-08-04
chore: use destructuring to get subject, predicate, object graph

jeswr:


TallTed:
It seems likely to prevent some future confusion if a comma can be added to the title of this PR, changing from `predicate, object graph` to `predicate, object, graph`.

PR #418, opened 2024-08-05
perf: use set

jeswr:
Uses a set rather than object for the final layer of the index.

Short term performance gain from being able to use built-in set operations.

Long term gain from being able to use symbols for indicies

First tests seem to indicate a performance **degredation**

main
```bash
$ node perf/N3Store-perf.js 
N3Store performance test
- Adding 16777216 triples to the default graph: 23.328s
* Memory usage for triples: 855MB
```

This branch

```bash
$ node perf/N3Store-perf.js 
N3Store performance test
- Adding 16777216 triples to the default graph: 39.701s
* Memory usage for triples: 1083MB
```

### DIFF
--- a/src/N3Reasoner.js
+++ b/src/N3Reasoner.js
@@ -39,7 +39,7 @@ export default class N3Reasoner {
           if (index2 = index1[value]) {
             if (v1) val1.value = Number(value);
             v2 = !(value = val2.value);
-            for (value in v2 ? index2 : { [value]: index2[value] }) {
+            for (value of (v2 ? index2 : [value])) {
               if (v2) val2.value = Number(value);
 
               if (i === rule.premise.length - 1)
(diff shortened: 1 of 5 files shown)


RubenVerborgh:
RE: degradation, know that the benchmark might have a flawed design. It's an atypical insertion pattern. Retrieval is typically much more common than insertion.

Could be an option to have a switchable implementation where different read/write performance balance is desired.

jeswr:
Closing this experiment. The numbers in the description already showed a clear insert-path regression (23.3s to 39.7s, 855MB to 1083MB) and I never followed up with read-path benchmarks to offset it. Since then the branch has also been overtaken by the index-level set operations on main (#438, #443, #451, #458, and the in-flight #601): those all walk the index with `for...in` / `key in` assuming plain-object leaves, so a Set leaf layer would now silently break intersection/difference even though git reports no conflicts. `Set.prototype.isSubsetOf` also needs Node 22+, which conflicts with our `node >=12` support. If we revisit Set-based leaves it should be a fresh branch with read-heavy benchmarks alongside the insert ones — the conditions for that are now recorded in #620.

PR #423, opened 2024-08-24
Feature: callback and event for comments

pietercolpaert:
For my project I need access to the comments in a turtle (etc.) file.

This pull request:
 1. extends the N3Parser.parse function with an optional commentCallback. If set, the lexer’s configuration will be overridden to also provide comment tokens.
 2. extends the N3StreamParser with a "comment" event in a separate commit

### DIFF
--- a/README.md
+++ b/README.md
@@ -105,7 +105,7 @@ the callback is invoked one last time with `null` for `quad`
 and a hash of prefixes as third argument.
 <br>
 
-Alternatively, an object may be supplied, where `onQuad`, `onPrefix` and `onComment` may be used to listen for `quads`, `prefixes` and `comments` as follows:
+Alternatively, an object can be supplied, where `onQuad`, `onPrefix` and `onComment` are used to listen for `quads`, `prefixes` and `comments` as follows:
 ```JavaScript
 const parser = new N3.Parser();
 
@@ -113,14 +113,14 @@ parser.parse(tomAndJerry, {
   // onQuad (required) accepts a listener of type (quad: RDF.Quad) => void
   onQuad: (err, quad) => { console.log(quad); },
   // onPrefix (optional) accepts a listener of type (prefix: string, iri: NamedNode) => void
-  onPrefix: (prefix, iri) => { console.log(prefix, 'expands to', iri); },
+  onPrefix: (prefix, iri) => { console.log(prefix, 'expands to', iri.value); },
   // onComment (optional) accepts a listener of type (comment: string) => void
   onComment: (comment) => { console.log('#', comment); },
 });
 ```
 
 <br>
-If no callbacks are provided, parsing happens synchronously returning an array of parsed quads.
+If no callbacks are provided, parsing happens synchronously returning an array of quads.
 
 ```JavaScript
 const parser = new N3.Parser();
@@ -193,7 +193,7 @@ function SlowConsumer() {
 
 A dedicated `prefix` event signals every prefix with `prefix` and `term` arguments.
 
-A dedicated `comment` event may be enabled by setting `comments: true` in the N3.StreamParser constructor.
+A dedicated `comment` event can be enabled by setting `comments: true` in the N3.StreamParser constructor.
 
 ## Writing
 
--- a/README.md
+++ b/README.md
@@ -89,19 +89,22 @@ parser.parse(
    c:Tom a c:Cat.
    c:Jerry a c:Mouse;
            c:smarterThan c:Tom.`,
-  (error, quad, prefixes) => {
+  (error, quad) => {
     if (quad)
       console.log(quad);
     else
-      console.log("# That's all, folks!", prefixes);
+      console.log("# That's all, folks!");
   });
 ```
 The callback's first argument is an optional error value, the second is a quad.
 If there are no more quads,
-the callback is invoked one last time with `null` for `quad`
-and a hash of prefixes as third argument.
+the callback is invoked one last time with `null` for `quad`.
 <br>
-Pass a second callback to `parse` to retrieve prefixes as they are read.
+In case you would also like to process prefixes, you can instead pass an object containing multiple callbacks.
+The callback to retrieve the quads is called `onQuad`.
+The callback to also retrieve prefixes as they are read is called `onPrefix`.
+The first argument is the prefix, the second is the IRI.
+There is also a third callback called `onComment` taking only one `comment` argument.
 <br>
 If no callbacks are provided, parsing happens synchronously.
 
@@ -168,6 +171,8 @@ function SlowConsumer() {
 ```
 
 A dedicated `prefix` event signals every prefix with `prefix` and `term` arguments.
+
+Also a `comment` event can be enabled through the options object of the N3.StreamParser constructor using: `comments: true`.
 
 ## Writing
 
--- a/README.md
+++ b/README.md
@@ -89,16 +89,17 @@ parser.parse(
    c:Tom a c:Cat.
    c:Jerry a c:Mouse;
            c:smarterThan c:Tom.`,
-  (error, quad) => {
+  (error, quad, prefixes) => {
     if (quad)
       console.log(quad);
     else
-      console.log("# That's all, folks!");
+      console.log("# That's all, folks!", prefixes);
   });
 ```
 The callback's first argument is an optional error value, the second is a quad.
 If there are no more quads,
-the callback is invoked one last time with `null` for `quad`.
+the callback is invoked one last time with `null` for `quad`
+and a hash of prefixes as third argument.
 <br>
 In case you would also like to process prefixes, you can instead pass an object containing multiple callbacks.
 The callback to retrieve the quads is called `onQuad`.
--- a/README.md
+++ b/README.md
@@ -83,12 +83,15 @@ we assume that a quad is simply a triple in a named or default graph.
 
 `N3.Parser` transforms Turtle, TriG, N-Triples, or N-Quads document into quads through a callback:
 ```JavaScript
+const tomAndJerry = `PREFIX c: <http://example.org/cartoons#>
+  # Tom is a cat
+  c:Tom a c:Cat.
+  c:Jerry a c:Mouse;
+    c:smarterThan c:Tom.`
+
 const parser = new N3.Parser();
-parser.parse(
-  `PREFIX c: <http://example.org/cartoons#>
-   c:Tom a c:Cat.
-   c:Jerry a c:Mouse;
-           c:smarterThan c:Tom.`,
+
+parser.parse(tomAndJerry,
   (error, quad, prefixes) => {
     if (quad)
       console.log(quad);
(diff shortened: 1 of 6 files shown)


src/N3Parser.js
```diff
@@ -1010,8 +1010,8 @@ export default class N3Parser {
 
   // ## Public methods
 
-  // ### `parse` parses the N3 input and emits each parsed quad through the callback
-  parse(input, quadCallback, prefixCallback) {
+  // ### `parse` parses the N3 input and emits each parsed quad through the callback. Optionally you can set a prefixCallback and a commentCallback
+  parse(input, quadCallback, prefixCallback, commentCallback) {
```
RubenVerborgh, line 1014: We're getting too many arguments here; let's support:

- `parse(input, onQuad)`
- `parse(input, onQuad, onPrefix)`
- `parse(input, { onQuad, onPrefix, onComment })`

the first two of which are for legacy compatibility.

src/N3Parser.js
```diff
@@ -1023,6 +1023,9 @@ export default class N3Parser {
     this._inversePredicate = false;
     this._quantified = Object.create(null);
 
+    // If the parser is requested to provide comments through a callback, override the current lexer’s configuration parameter to emit comments
+    if (commentCallback) this._lexer._comments = true;
```
RubenVerborgh, line 1027: Shouldn't set private (`_`) fields of others; need to set this up differently (constructor).

src/N3Parser.js
```diff
@@ -1040,8 +1043,16 @@ export default class N3Parser {
     this._lexer.tokenize(input, (error, token) => {
       if (error !== null)
         this._callback(error), this._callback = noop;
-      else if (this._readCallback)
-        this._readCallback = this._readCallback(token);
+      else if (this._readCallback) {
+        if (token.type === 'comment') {
```
RubenVerborgh, line 1047: Expecting significant performance implication of checking this on every token; will likely require different approach.

RubenVerborgh (commented):
Thanks, Pieter!
Some quick highlights; should definitely check performance impact.
We can likely omit most of the code comments.

src/N3Parser.js
pietercolpaert: In the constructor might not be the right place though: only when the parse function has a commentcallback you want to enable the comments in the lexer. I would propose a setter then instead. Do you agree?

src/N3Parser.js
pietercolpaert: Ok, will amend my PR accordingly

src/N3Parser.js
pietercolpaert: Another parse function, maybe called `parseWithComments`, would mitigate this I guess. Would that be acceptable?

src/N3Parser.js
RubenVerborgh: Setter is good 👍

src/N3Parser.js
RubenVerborgh: I'm thinking something akin to:
```
{
//...
  _processNextToken(token) {
    this._readCallback = this._readCallback(token);
  }
//...
}
```
which, in the case of comments, is monkey-patched with
```
this._processNextToken = function (token) {
  if (token.type === 'comment')
    this._commentCallback(token.value);
  else
    this._readCallback = this._readCallback(token);
}
```
or similar.

Possibly instead of monkey-patching, it _could_ be a boolean check in `_processNextToken` for whether or not comment mode is enabled.

src/N3Parser.js
RubenVerborgh: Alternatively, pass `comment` option to the constructor, which then initiates monkey patching (see below). And then `this._commentCallback` is just set to `() => null` if no `commentCallback` is passed.

RubenVerborgh:
Note that the slightly weird suggestions here are because this parser is written quite close to the V8 engine; some functions execute 200,000+ times per second, so every `if` and `===` matter.

src/N3Parser.js
pietercolpaert: I am afraid developers will find it counter-intuitive that they must add a callback, _and must_ also have constructed the parser in a different way or it won’t work. I think the setter is the best solution here.

src/N3Parser.js
pietercolpaert: I’ve implemented monkey-patching

src/N3Parser.js
pietercolpaert: Implemented the setter

pietercolpaert:
All comments have been processed. At your own convenience, can you give this PR another review?

src/N3Lexer.js
```diff
@@ -463,6 +463,11 @@ export default class N3Lexer {
 
   // ## Public methods
 
+  // ### `setComments` enables or disables creating a token per comment using a boolean.
+  setComments(enable) {
+    this._comments = enable;
+  }
+
```
RubenVerborgh, line 470: FYI, JS can do setters by just

```js
{
//...
  set comments(value) {
    this._comments = value;
  }
//...
}
```

but in this case, we can probably just rename `_comments` into `comments` to make it public.

src/N3Parser.js
```diff
@@ -1035,14 +1047,36 @@ export default class N3Parser {
...
-    this._callback = quadCallback;
-    this._lexer.tokenize(input, (error, token) => {
+    let processNextToken = (error, token) => {
       if (error !== null)
         this._callback(error), this._callback = noop;
-      else if (this._readCallback)
+      else if (this._readCallback) {
         this._readCallback = this._readCallback(token);
-    });
+      }
```
RubenVerborgh, line 1055: Let's keep the `{…}` off for single lines (the linter should normally complain, weird).

src/N3Parser.js
```diff
@@ -1035,14 +1047,36 @@ export default class N3Parser {
...
+      processNextToken = (error, token) => {
+        if (error !== null)
+          this._callback(error), this._callback = noop;
+        else if (this._readCallback) {
+          if (token.type === 'comment') {
+            onComment(token.value);
+          }
+          else {
+            this._readCallback = this._readCallback(token);
+          }
```
RubenVerborgh, line 1072:
```suggestion
          if (token.type === 'comment')
            onComment(token.value);
          else
            this._readCallback = this._readCallback(token);
```

src/N3StreamParser.js
```diff
@@ -18,12 +18,13 @@ export default class N3StreamParser extends Transform {
...
-    },
+    }, {
       // Handle quads by pushing them down the pipeline
-      (error, quad) => { error && this.emit('error', error) || quad && this.push(quad); },
+      onQuad: (error, quad) => { error && this.emit('error', error) || quad && this.push(quad); },
       // Emit prefixes through the `prefix` event
-      (prefix, uri) => { this.emit('prefix', prefix, uri); },
-    );
+      onPrefix: (prefix, uri) => { this.emit('prefix', prefix, uri); },
+      onComment: comment => { this.emit('comment', comment); },
```
RubenVerborgh, line 26: This one is unconditional though, so there will be performance implications. (That's where a constructor param would perhaps help.)

RubenVerborgh (commented):
The CI is not kicking in, weird.

Mostly okay, but we'll need the result from the performance tests (they're in the repo). What is the difference? The streaming parser has unconditional comments currently, so there would be a penalty (but the perf tests aren't looking for them unfortunately).

src/N3Lexer.js
pietercolpaert: Just making it public makes sense. Done that

src/N3Parser.js
pietercolpaert: Fixed

src/N3Parser.js
pietercolpaert: Applied this in my manual commit

src/N3StreamParser.js
pietercolpaert: This will now not happen by default anymore: a constructor parameter is required now

pietercolpaert:
I made emitting comments in the streamparser optional and not the default now.

test/N3Parser-test.js
```diff
@@ -41,6 +41,20 @@ describe('Parser', () => {
...
+    it(
+      'should parse three triples with comments if no comment callback is set',
+      shouldParse('<a> <b> #comment2\n <c> . \n<d> <e> <f>.\n<g> <h> <i>.',
+                  ['a', 'b', 'c'],
+                  ['d', 'e', 'f'],
+                  ['g', 'h', 'i']),
+    );
+
+    it(
+      'should callback comments when a comment allback is set',
```
RubenVerborgh, line 53:
```suggestion
      'should callback comments when a comment callback is set',
```

RubenVerborgh (approved):
Thanks—I think I'm happy, I'll just have another pair of eyes on this!

jeswr (approved):
LGTM

pietercolpaert:
Thanks @jeswr and @RubenVerborgh for the extremely fast and thorough review process!

I’ve added some tests just now that fix the coverage. It appeared locally the coverage didn’t trigger due to other experimental tests I was running that did cover the line.

Squashing is something @RubenVerborgh will do?

Should I also add documentation in the README.md on how to enable comments?

jeswr:
> Should I also add documentation in the README.md on how to enable comments?

Yes, that would be a good idea! Ruben and I both have permission to merge so can do that once the new documentation is added.

pietercolpaert:
Added the documentation in the README.md, finalizing this PR!

README.md
```diff
@@ -89,19 +89,22 @@ parser.parse(
...
-the callback is invoked one last time with `null` for `quad`
-and a hash of prefixes as third argument.
+the callback is invoked one last time with `null` for `quad`.
 <br>
-Pass a second callback to `parse` to retrieve prefixes as they are read.
+In case you would also like to process prefixes, you can instead pass an object containing multiple callbacks.
+The callback to retrieve the quads is called `onQuad`.
+The callback to also retrieve prefixes as they are read is called `onPrefix`.
+The first argument is the prefix, the second is the IRI.
+There is also a third callback called `onComment` taking only one `comment` argument.
```
jeswr, line 107: Thanks for correcting this. For completeness could you add examples as these are usually easier for developers to parse than text descriptions.

Please also note that `onComment` is available when `comment: true` is set and include this in the example.

README.md
```diff
@@ -89,19 +89,22 @@ parser.parse(
    c:Tom a c:Cat.
    c:Jerry a c:Mouse;
            c:smarterThan c:Tom.`,
-  (error, quad, prefixes) => {
+  (error, quad) => {
     if (quad)
       console.log(quad);
     else
-      console.log("# That's all, folks!", prefixes);
```
jeswr, line 96: When was this hash functionality as part of the callback deprecated?

README.md
pietercolpaert: onComment is enabled from the moment you set the callback. No need for another constructor parameter here. This is different in the StreamParser.

I’m afraid the examples will get a bit much. The simplest example is written out in code. For more advanced features (prefixes and comments) you need to read the text?

README.md
pietercolpaert: It wasn’t, but I interpreted @RubenVerborgh’s text above as this would be deprecated, but I see I could be wrong. I’ll re-add this!

README.md
jeswr: No need to worry, I've done some cleanup :)

jeswr:
Thanks for your work @pietercolpaert !

RubenVerborgh:
Limitation: timing of `comment` event in `StreamParser` is unreliable as it is unsynchronized. https://github.com/rdfjs/N3.js/issues/440

PR #443, opened 2024-09-07
perf: perform matching over indices

jeswr:
When creating a new dataset with `#match` do this over the indices rather than requiring the RDF/JS layer

## Summary by CodeRabbit

- **New Features**
	- Introduced a new function for advanced recursive matching in the dataset processing, enhancing the ability to match identifiers within store indexes.

- **Bug Fixes**
	- Improved error handling when translating IRIs into internal IDs, ensuring more reliable dataset processing.

- **Tests**
	- Expanded test suite to include precise matching checks for dataset quads, increasing the robustness and reliability of the tests.

### DIFF
--- a/src/N3Store.js
+++ b/src/N3Store.js
@@ -988,6 +988,26 @@ export default class N3Store {
 }
 
 /**
+ * Performs matching over the store indexes
+ */
+function indexMatch(index, ids, depth = 0) {
+  const ind = ids[depth];
+  if (ind && !(ind in index))
+    return false;
+
+  let target = false;
+  for (const key in (ind ? { [ind]: index[ind] } : index)) {
+    const result = depth === 2 ? null : indexMatch(index[key], ids, depth + 1);
+
+    if (result !== false) {
+      target = target || Object.create(null);
+      target[key] = result;
+    }
+  }
+  return target;
+}
+
+/**
  * A class that implements both DatasetCore and Readable.
  */
 class DatasetCoreAndReadableStream extends Readable {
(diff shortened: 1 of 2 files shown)


src/N3Store.js
```diff
@@ -1026,6 +1026,26 @@ export default class N3Store {
   }
 }
 
+/**
+ * Performs matching over the store indexes
+ */
```
jeswr, line 1031: Shoud add similar docs to #438 around falsy values

src/N3Store.js
```diff
@@ -1039,8 +1059,27 @@ class DatasetCoreAndReadableStream extends Readable {
...
+          object    && !(objectId    = newStore._termToNumericId(object)))
+        return newStore;
+
+      for (const graph in graphs) {
+        const subjects = indexMatch(graphs[graph].subjects, [subjectId, predicateId, objectId]);
+        if (subjects) {
+          newStore._graphs[graph] = {
+            subjects,
+            predicates: indexMatch(graphs[graph].predicates, [predicateId, objectId, subjectId]),
+            objects: indexMatch(graphs[graph].objects, [objectId, subjectId, predicateId]),
```
jeswr, line 1078: We could optimise slightly further by using the same technique as `#match` to determine which index to match over first as this will be more efficient for identifying if there are no matches. But it would be sufficient to leave that work for a follow up PR.

src/N3Store.js
jeswr, line 1031:
```suggestion
/**
 * Returns a subset of the `index` with that part of the index
 * matching the `ids` array. `ids` contains 3 elements that are
 * either numerical ids; or `null`.
 *
 * s1 and s2 *must* belong to Stores that share an `_entityIndex`.
 *
 * `false` is returned when there are no matching indices; this should
 * *not* be set as the value for an index.
 */
```

PR #467, opened 2024-10-30
fix: Correctly handle custom data factories by binding functions

### LINKED ISSUE #466: Custom data factories are incorrectly handled
When injecting a custom factory into the `StreamParser` (possibly other places as well), the factory methods are being used as standalone functions instead of methods on the factory object.
This makes it so that custom factories (such as `rdf-data-factory`) that rely on an internal `this` state will start failing, which causes issues such as https://github.com/rubensworks/rdf-dereference.js/issues/55.

I suspect the solution may be as simple as just `bind`-ing the extracted factory functions to `factory` [here](https://github.com/rdfjs/N3.js/blob/main/src/N3Parser.js#L1086-L1092).

smessie:
Closes #466.
Tested this and this seems to resolve https://github.com/rubensworks/rdf-dereference.js/issues/55

## Summary by CodeRabbit

- **New Features**
	- Enhanced the parser's interaction with the data factory for improved functionality.
	- Improved method bindings to ensure correct context during execution.
	- Added new test cases to validate the parser's handling of RDF-star syntax and complex scenarios, including edge cases.

- **Bug Fixes**
	- Resolved issues related to method scoping in the parser, ensuring expected behavior during data processing.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -11088,21 +11088,6 @@
       "integrity": "sha512-/aCDEGatGvZ2BIk+HmLf4ifCJFwvKFNb9/JeZPMulfgFracn9QFcAf5GO8B/mweUjSoblS5In0cWhqpfs/5PQA==",
       "dev": true
     },
-    "node_modules/typescript": {
-      "version": "5.6.2",
-      "resolved": "https://registry.npmjs.org/typescript/-/typescript-5.6.2.tgz",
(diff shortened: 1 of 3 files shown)


jeswr:
That's @smessie ! Could you add some regression tests?

smessie:
Added a small test @jeswr

RubenVerborgh (changes requested):
Instead of binding, can we call them on the `factory` object?
These functions are called hundreds of thousands of times per second; the difference might matter in some engines.

jeswr (approved):

jeswr:
I did performance testing with the latest commit and there is no statistically significant difference compared to the main branch.

PR #484, opened 2025-01-14
Add support for base directions

rubensworks:
This adds support for the new base directions in literals for RDF 1.2 (data factory, parsing, serializing): https://w3c.github.io/rdf-concepts/spec/#dfn-dir-lang-string
This follows the RDF/JS data model: https://rdf.js.org/data-model-spec/#dom-literal-direction

Unless I made some mistake, this is backwards-compatible, so no major version change is needed for this one.
I think it's safe to release this already, as the new triple terms for RDF 1.2 _will_ be a breaking change (as it will break RDF-star support).

## Release Notes

- **New Features**
  - Added support for language direction in RDF literals
  - Introduced ability to specify text direction (left-to-right or right-to-left) for language-tagged literals

- **Improvements**
  - Enhanced parsing, lexing, and serialization of literals with directional language tags
  - Updated internal handling of literal representations to support language and direction

- **Testing**
  - Expanded test coverage for new language direction functionality
  - Added comprehensive tests for parsing, serializing, and handling directional literals

### DIFF
--- a/src/N3Writer.js
+++ b/src/N3Writer.js
@@ -172,8 +172,9 @@ export default class N3Writer {
       value = value.replace(escapeAll, characterReplacer);
 
     // Write a language-tagged literal
+    const direction = literal.direction ? `--${literal.direction}` : '';
     if (literal.language)
-      return `"${value}"@${literal.language}`;
+      return `"${value}"@${literal.language}${direction}`;
 
     // Write dedicated literals per data type
     if (this._lineMode) {
--- a/test/N3Writer-test.js
+++ b/test/N3Writer-test.js
@@ -76,6 +76,12 @@ describe('Writer', () => {
                       '<a> <b> "cde"@en-us.\n'),
     );
 
+    it(
+        'should serialize a literal with a language and direction',
+        shouldSerialize(['a', 'b', '"cde"@en-us--ltr'],
+            '<a> <b> "cde"@en-us--ltr.\n'),
+    );
+
     // e.g. http://vocab.getty.edu/aat/300264727.ttl
     it(
       'should serialize a literal with an artificial language',


RubenVerborgh:
Thanks, @rubensworks, great to have this.

> Unless I made some mistake, this is backwards-compatible, so no major version change is needed for this one.
> I think it's safe to release this already,

We just need to be aware that N3.js is also a Turtle (1.1) validator. So we'd need to either:
- No semver.major: Introduce this new behavior behind a 1.2 flag.
  - The default parser can remain permissive (as it currently is) and thus accept 1.2 bits of syntax.
  - Setting the format to `text/turtle` should however remain restrictive and thus reject this new syntax.
- semver.major: Make the 1.2 syntax the default for Turtle.

However, I don't think the second one would even be an option, since I presume that the MIME type will no longer be `text/turtle`?

My quick conclusion:
- Needs a new 1.2 flag (with the new Turtle 1.2 content type? TriG as well? N-Triples?)
- The existing strict Turtle/TriG/… modes need to reject this new syntax, unless semver.major.

rubensworks:
Ok, if we want consider usage of base directions in 1.1 to throw, then we'd need a new mode indeed.

> since I presume that the MIME type will no longer be text/turtle?

The media type will actually remain the same.

> semver.major: Make the 1.2 syntax the default for Turtle.

My feeling is that a major release may be a more feasible option here.
The flag option would complicate the new triple terms syntax (not implemented yet) by a lot, since there's also the rdfStar flag to take into account.
The next major release could then also remove the rdfStar flag all together, since it will be superseded by RDF 1.2.

So what if we put this PR on hold (or merge it into a `next/major` branch, for which a prerelease could be published), and only merge and (semver) release it once the new triple terms are in?

RubenVerborgh:
> The media type will actually remain the same.

That… is something we should look into.
I don't think the definition`text/turtle` can retroactively be changed in that way.

Yes, `text/html` has undergone changes, but it has a version mechanism inside.

Has there been any discussion on that? Any objections?

And how can a client distinguish between whether it should use a 1.1 or 1.2 parser on a document?

> My feeling is that a major release may be a more feasible option here.
> The flag option would complicate the new triple terms syntax (not implemented yet) by a lot, since there's also the rdfStar flag to take into account.

[As a background to the below: N3.js has the default mode (quirks mode), which parses a superset of several RDF syntaxes, and content-type-specific strict modes, which strictly parse the respective syntaxes.]

Not necessarily; the only flag consideration is that, in the current semver.major:
- default mode accepts everything it accepted before (and possibly more in addition)
- strict modes accept and reject the same set as before

So we can extend default mode as much as we want (unless there's a conflict, then we should keep existing behavior), and we can add as many strict modes as we want. And we're not obliged to support `rdfStar` in any of the new strict modes we add.

Concretely, for base directions:
- Default mode can happily accept them.
- A new strict mode (or a flag to an existing strict mode) can happily accept them.

The only thing we _can't_ do is change the set of an existing strict mode without a flag.

> The next major release could then also remove the rdfStar flag all together, since it will be superseded by RDF 1.2.

Agreed.

rubensworks:
> I don't think the definitiontext/turtle can retroactively be changed in that way.

I think the reasoning here is that there are no breaking changes going from 1.1 to 1.2.
Only things that were not allowed before, now become allowed.
Similar to going from 1.0 to 1.1.

> Has there been any discussion on that? Any objections?

No discussions or objections AFAIK. I also don't really object to it myself 😅 
But if you feel strongly about it, I guess you could raise an issue: https://github.com/w3c/rdf-star-wg/issues

> The only thing we can't do is change the set of an existing strict mode without a flag.

Within the current major version range that definitely makes sense.
Related do this, would you want to keep 1.1 support in the next major version (where 1.2 features are rejected)?

RubenVerborgh:
> Related do this, would you want to keep 1.1 support in the next major version (where 1.2 features are rejected)?

That depends on the answer of the first point.

> I think the reasoning here is that there are no breaking changes going from 1.1 to 1.2.
> Only things that were not allowed before, now become allowed.

It depends on how one defines "breaking change".

A valid Turtle parser as per the 1.1 spec is a parser that 1) creates the correct syntax tree for any valid Turtle 1.1 document, 2) rejects everything else. (There are rejection tests within the RDF suite in general.)

So that would be breaking.

> But if you feel strongly about it, I guess you could raise an issue: https://github.com/w3c/rdf-star-wg/issues

I'll probably need to indeed.

rubensworks:
> A valid Turtle parser as per the 1.1 spec is a parser that 1) creates the correct syntax tree for any valid Turtle 1.1 document, 2) rejects everything else. (There are rejection tests within the RDF suite in general.)

Based on my understanding of the [conformance section](https://www.w3.org/TR/turtle/#h2_conformance), 1 is true indeed, but I'm not sure about 2 (unless that is written elsewhere).

In any case, all of the rejection tests from 1.1 still apply in 1.2.

But it may be good to raise this as an issue to the WG in any case. It may be good to have the WG formally state the intentions.

RubenVerborgh:
Issue raised at https://github.com/w3c/rdf-star-wg/issues/141

Rejection tests are exemplary only; they are not the full range of rejected output.

PR #500, opened 2025-03-03
feat: add log:isImpliedBy feature flag

jeswr:


src/N3Lexer.js
```diff
@@ -54,6 +54,9 @@ export default class N3Lexer {
     this._endOfFile = /^(?:#[^\n\r]*)?$/;
     options = options || {};
 
+    // Whether the log:isImpliedBy predicate is supported
+    this._isImpliedBy = options.isImpliedBy;
+
```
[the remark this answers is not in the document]
jeswr: good bot - now added the docs

README.md
```diff
@@ -159,6 +159,13 @@ prefix or, if set to an empty string, completely disable prefixing:
 const parser = new N3.Parser({ blankNodePrefix: '' });
 ```
 
+By default, `N3.Parser` does not use the `log:isImpliedBy` predicate when parsing
+backwards chaining rules in Notation3. To enable parsing with this predicate use
```
jeswr, line 163:
```suggestion
backwards chaining rules in Notation3. To enable parsing with this predicate, use 
```

README.md
```diff
@@ -159,6 +159,13 @@ prefix or, if set to an empty string, completely disable prefixing:
 const parser = new N3.Parser({ blankNodePrefix: '' });
 ```
 
+By default, `N3.Parser` does not use the `log:isImpliedBy` predicate when parsing
```
RubenVerborgh, line 162: Maybe describe the default behavior?
Something like:

The parser can output a backwards chaining rule such as `_:q <= _:p.` in two ways:
- as `_:p log:implies _:q.` (default)
- as `_:q log:isImpliedBy _:p.` (when the `isImpliedBy` flag is  set to `true`)

src/N3Parser.js
```diff
@@ -27,14 +27,16 @@ export default class N3Parser {
...
+    this._isImpliedBy = options.isImpliedBy;
     // Support nesting of triples
     this._supportsRDFStar = format === '' || /star|\*$/.test(format);
     // Disable relative IRIs in N-Triples or N-Quads mode
     if (isLineMode)
       this._resolveRelativeIRI = iri => { return null; };
     this._blankNodePrefix = typeof options.blankNodePrefix !== 'string' ? '' :
                               options.blankNodePrefix.replace(/^(?!_:)/, '_:');
-    this._lexer = options.lexer || new N3Lexer({ lineMode: isLineMode, n3: isN3 });
+    this._lexer = options.lexer || new N3Lexer({ lineMode: isLineMode, n3: isN3, isImpliedBy: this._isImpliedBy });
```
RubenVerborgh, line 39: Suggestion: make the lexer do this by default? I.e., change lexer behavior?

test/N3Parser-test.js
```diff
@@ -1983,6 +1983,7 @@ describe('Parser', () => {
 
   describe('A Parser instance for the N3 format', () => {
     function parser() { return new Parser({ baseIRI: BASE_IRI, format: 'N3' }); }
+    function parserIsImpledBy() { return new Parser({ baseIRI: BASE_IRI, format: 'N3', isImpliedBy: true }); }
```
RubenVerborgh, line 1986:
```suggestion
    function parserIsImpliedBy() { return new Parser({ baseIRI: BASE_IRI, format: 'N3', isImpliedBy: true }); }
```

RubenVerborgh (approved):

src/N3Parser.js
jeswr: To clarify, is the suggestion that the lexer *always* returns the `abbreviation` type and then the parser decides whether to produce `log:implies` or `log:isImpliedBy`.

If so, can be done, but will be slightly messier to implement than this option.

RubenVerborgh: Ah no, keep whatever option is less messy. Looks good in the lexer the way it is!

test/N3Parser-test.js
jeswr: good catch

PR #501, opened 2025-03-03
feat: make return quads return a boolean indicating if all removed quads were in store

### LINKED ISSUE #479: The `N3Store.removeQuads` function silently fails when a quad is not successfully removed
The [`N3Store.removeQuads`](https://github.com/rdfjs/N3.js/blob/4b54ae42e00ce28606dd0b853c5b822545de8d8c/src/N3Store.js#L449) function is not reporting any error when a quad is not successfully removed from the store.

This function just calls the [`N3Store.removeQuad`](https://github.com/rdfjs/N3.js/blob/4b54ae42e00ce28606dd0b853c5b822545de8d8c/src/N3Store.js#L451) function without taking into account its return value (true/false).

The `N3Store.removeQuads` function should throw an error or return something to inform that the operation failed partially or totally.

Maybe this function should check that all the quads exists in the store before to delete them.

jeswr:
Closes #479 

Technically a breaking change, may not be the right design choice. Lets discuss further in issue first.

### DIFF
--- a/src/N3Store.js
+++ b/src/N3Store.js
@@ -446,9 +446,12 @@ export default class N3Store {
   }
 
   // ### `removeQuads` removes multiple quads from the store
+  // returns `true` if all quads were removed, `false` if some were not found
   removeQuads(quads) {
+    let removed = quads.length >= 0;
     for (let i = 0; i < quads.length; i++)
-      this.removeQuad(quads[i]);
+      removed = this.removeQuad(quads[i]) && removed;
+    return removed;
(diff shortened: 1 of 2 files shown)


jeswr:
Since the return type changes from `undefined` to `boolean`, treating this as technically breaking: labelling `semver.major` and parking it to be picked up as part of the next major release (together with the rebase, the `let removed = true;` simplification, and an empty-array test). #479 stays open until then.

PR #503, opened 2025-03-07
fix: allow import to be treated as `Promise<Store>` per the RDF/JS spec

jeswr:
Uses a `Proxy` so that the return of `import` has the type `EventEmitter & Promise<Store & EventEmitter>` making it compatible both with the:
 - Existing return type of `import` (`EventEmitter`)
 - RDF/JS return type of `import` (`Promise<Store>`)
 
 cc @RubenVerborgh 

## Summary by CodeRabbit

- **New Features**
	- Introduced a new dependency for enhanced event handling.
	- Improved the asynchronous data import process with robust event handling and promise resolution.
- **Tests**
	- Expanded test coverage to validate seamless stream processing, proper event handling, and reliable error management.

### DIFF
--- a/package.json
+++ b/package.json
@@ -25,6 +25,7 @@
   ],
   "dependencies": {
     "buffer": "^6.0.3",
+    "events": "^3.3.0",
     "queue-microtask": "^1.1.2",
     "readable-stream": "^4.0.0"
   },
--- a/src/N3Store.js
+++ b/src/N3Store.js
@@ -411,8 +411,8 @@ export default class N3Store {
 
   /**
    * `import` adds a stream of quads to the store
-   * 
-   * @returns {EventEmitter & Promise<Store & EventEmitter>} A proxy object that acts as both an EventEmitter 
+   *
+   * @returns {EventEmitter & Promise<Store & EventEmitter>} A proxy object that acts as both an EventEmitter
    * (for backward compatibility) and a Promise that resolves to the store when the stream is complete.
    */
   import(stream) {
--- a/src/N3Store.js
+++ b/src/N3Store.js
@@ -1,4 +1,5 @@
 // **N3Store** objects store N3 quads by graph in memory.
+import EventEmitter from 'events';
 import { Readable } from 'readable-stream';
 import { default as N3DataFactory, termToId, termFromId } from './N3DataFactory';
 import namespaces from './IRIs';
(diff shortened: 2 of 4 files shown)


src/N3Store.js
```diff
@@ -410,8 +411,57 @@ export default class N3Store {
...
+        };
+
+          // eslint-disable-next-line func-style
+        const onError = err => {
+          stream.removeListener('end', onEnd);
+          reject(err);
+        };
+
+        stream.once('end', onEnd);
+        stream.once('error', onError);
```
jeswr, line 450: I wanted to only create this promise when `then`, `catch`, or `finally` were used in the below proxy - and thus only start listening to the stream at that point in time.


Unfortunately, that turned out to not be an option; as `EventEmitters` error if the error event is called and there are no error listeners - which causes the last added test case to break.

package.json
```diff
@@ -25,6 +25,7 @@
   ],
   "dependencies": {
     "buffer": "^6.0.3",
+    "events": "^3.3.0",
```
jeswr, line 28: This is already a transitive dependency via `readable-stream`

src/N3Store.js
```diff
@@ -410,8 +411,57 @@ export default class N3Store {
...
+
+    // Create a promise that resolves when the stream ends
+    const promise = new Promise((resolve, reject) => {
+      // Create proxy that combines N3Store with EventEmitter capabilities
+      const storeProxy = new Proxy(this, {
+        get(target, prop, receiver) {
+          if (prop in EventEmitter.prototype) {
+            return Reflect.get(stream, prop, receiver);
+          }
+          return Reflect.get(target, prop, receiver);
```
jeswr, line 425:
```suggestion
          return Reflect.get(prop in EventEmitter.prototype ? stream : target, prop, receiver);
```

src/N3Store.js
```diff
@@ -410,8 +411,54 @@ export default class N3Store {
 
   // ### `import` adds a stream of quads to the store
```
jeswr, line 412:
```suggestion
  /**
   * `import` adds a stream of quads to the store
   * 
   * @returns {EventEmitter & Promise<Store & EventEmitter>} A proxy object that acts as both an EventEmitter 
   * (for backward compatibility) and a Promise that resolves to the store when the stream is complete.
   */
```

src/N3Store.js
```diff
@@ -408,10 +409,61 @@ export default class N3Store {
     return !this.readQuads(subjectOrQuad, predicate, object, graph).next().done;
   }
 
-  // ### `import` adds a stream of quads to the store
+  /**
+   * `import` adds a stream of quads to the store
+   * 
+   * @returns {EventEmitter & Promise<Store & EventEmitter>} A proxy object that acts as both an EventEmitter 
+   * (for backward compatibility) and a Promise that resolves to the store when the stream is complete.
+   */
```
jeswr, line 417:
```suggestion
  /**
   * `import` adds a stream of quads to the store
   *
   * @returns {EventEmitter & Promise<Store & EventEmitter>} A proxy object that acts as both an EventEmitter
   * (for backward compatibility) and a Promise that resolves to the store when the stream is complete.
   */
```

jeswr:
Superceded by #636

PR #513, opened 2025-03-28
feat: Support writing relative IRIs with dot segments

### LINKED ISSUE #512: Generate relative parents when writing with baseIRI.
When supplying a baseIRI, the Writer might want follow the same algorithm as path.relative to also allow for relative parent IRIs.

```typescript
  const baseIRI = "https://example.com/ns/1/2/3/";
  const parser = new Parser({ baseIRI });

  const input = "<a> <../../b> <../..>.";
  const turtle = parser.parse(input);

  const writer = new Writer({ baseIRI });

  writer.addQuads(turtle);
  const txt = await new Promise((res) => writer.end((_, e) => res(e)));
  console.log("Expected\n" + input);
  console.log("Got\n" + txt);
```

```text
Expected
<a> <../../b> <../..>.
Got
<a> <https://example.com/ns/1/b> <https://example.com/ns/1/>.
```

Is this feature on the roadmap or should I stick with regexing the resulting output?

smessie:
Closes #512 

The Writer already supported relative IRIs, but only if they included the set `baseIRI`.
This PR rewrites how IRIs are relativized to allow dot segments (`../`) in the relative IRIs such that relative IRIs are also supported if the origin (and optionally a part of the path) of the baseIRI is included, where it then writes a relative IRI using dot segments.

## Summary by CodeRabbit

- **New Features**
  - Introduced a dedicated class for enhanced base IRI management, now available through our public interface.

- **Refactor**
  - Streamlined internal IRI processing logic for improved efficiency and reliability.

- **Tests**
  - Expanded the test suite with additional scenarios to validate proper IRI transformation, including various edge cases.

- **Chores**
  - Updated dependency management settings, allowing a higher number of module imports.


package.json
```diff
@@ -26,7 +26,8 @@
   "dependencies": {
     "buffer": "^6.0.3",
     "queue-microtask": "^1.1.2",
-    "readable-stream": "^4.0.0"
+    "readable-stream": "^4.0.0",
+    "relativize-url": "^0.1.0"
```
jeswr, line 30: Probably best to avoid dependency creep if possible - I assume that to write the relative URL functionality yourself isn't more than a few lines of code(?)

RubenVerborgh:
> This PR introduces to allow parent references (`../`) in the relative IRIs such that relative IRIs are also supported if the origin (and optionally a part of the path) of the baseIRI is included, where it then writes a relative IRI using parent references.

Apologies, I don't understand the case. Could you clarify with an example and walk me through it (preferably also one that demonstrates the need/relevance)? Is this usability (shorter output) or does this enable scenarios that don't work otherwise?

I wasn't able to fully reconstruct an understanding from the test (which might very well be me).

smessie:
> Apologies, I don't understand the case. Could you clarify with an example and walk me through it (preferably also one that demonstrates the need/relevance)? Is this usability (shorter output) or does this enable scenarios that don't work otherwise?
> 
> I wasn't able to fully reconstruct an understanding from the test (which might very well be me).

Does the example in the linked issue help?
The PR linked in that issue should also demonstrate the need/relevance.

Let me know if that would not suffice, then I'll see if I can work out an additional example.

RubenVerborgh:
> Does the example in the linked issue help?

Yes it does, I had missed that, sorry! Seems like a good one to have, and by default; I don't think it needs a switch. Working on a review atm.

src/N3Writer.js
```diff
@@ -153,7 +158,10 @@ export default class N3Writer {
     }
     let iri = entity.value;
     // Use relative IRIs if requested and possible
-    if (this._baseMatcher && this._baseMatcher.test(iri))
+    if (this._baseOriginMatcher && this._baseOriginMatcher.test(iri)) {
+      iri = this._relater.relate(iri);
```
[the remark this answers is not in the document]
RubenVerborgh: Yeah I checked the code, it's all `URL`-based. We can't use it for RDF.

src/N3Writer.js
```diff
@@ -61,6 +62,10 @@ export default class N3Writer {
         this._baseMatcher = new RegExp(`^${escapeRegex(options.baseIRI)
             }${options.baseIRI.endsWith('/') ? '' : '[#?]'}`);
         this._baseLength = options.baseIRI.length;
+        if (options.allowParentReferences) {
```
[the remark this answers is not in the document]
smessie: The changes in this PR go one step further than relative URIs though.
The functionality on main (without this PR) already support relative URIs (when `baseIRI` is set), but only when the `baseIRI` is fully included in the URI, so only if the URI is deeper than the baseIRI; see https://github.com/rdfjs/N3.js/blob/fe20ee377e577b2dda692517a0438ff403c65f33/src/N3Writer.js#L155-L157

That's why I chose for a more specific option name which signal better that it is a special version of supporting relative URIs where also parent references (`../`) are allowed to be used.

I can change the option to `relativeUris`, but I think that will not signal the actual behaviour and will lead to confusion.

RubenVerborgh (changes requested):
Thanks, @smessie. It's a good feature to have.
However, the implementation cannot depend on `URL`, as it performs certain normalizations that are equivalent in HTTP URL-land, but not in RDF land.

A possible algorithm is to develop a regex that decomposes the relative URL in different parts. For example (non-escaped for clarity) `^http://example.org/(foo/(bar/(baz/)))` and then, depending on match length, substitute as planned. The substitutions could even be precalculated, such that (I haven't actually counted): `{ 20: '../../', 26: '../' }`, etc. The precalculations can be interesting, since the parser is reusable and should be as fast as possible in the streaming bit. Other algos work too, of course; the regex engine just tends to be very fast.

src/N3Writer.js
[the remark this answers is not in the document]
RubenVerborgh: I think (in hindsight) it was an incomplete implementation originally to not consider the entire base IRI range. I don't see reasons to not do it by default as part of the existing base IRI functionality. But perhaps it would be good if we find counterexamples where such behavior would not be desired.

src/N3Writer.js
```diff
@@ -58,9 +58,30 @@ export default class N3Writer {
       this._prefixIRIs = Object.create(null);
       options.prefixes && this.addPrefixes(options.prefixes);
       if (options.baseIRI) {
-        this._baseMatcher = new RegExp(`^${escapeRegex(options.baseIRI)
-            }${options.baseIRI.endsWith('/') ? '' : '[#?]'}`);
+        const escapedBaseIRI = escapeRegex(options.baseIRI) + (options.baseIRI.endsWith('/') ? '' : '[#?]');
```
RubenVerborgh, line 61: We'll need to perform some sanitization on the base IRI. A couple of breakage points I see (and there might be more), is with `baseIRI` set to:

- `""` (well the `if` will fail then, so that's good)
- `/`
- `//`
- `//////`
- `http://example.org//` (I don't even know the answer to this one)
- `http://example.org///////////////` (idem)
- `http://example.org/./`
- `http://example.org/.` 
- `http://example.org/../`
- `http://example.org/..`
- `http://example.org/../../../../../../`
- `http://example.org/../../../../../..`
- `http://example.org/../../../foo/../../`
- `http://example.org/?`
- `http://example.org/#`
- `http://example.org/#?`
- `http://example.org/?#`
- `http://example.org/?/`
- `http://example.org/#/`
- `#`
- `?`

So it looks like we'll probably want some dedicated test automation for this, i.e., basically a function we can call such that:

```javascript
checkRelativizes({ base: 'http://example.org/////foo' , input: 'http://example.org/////foo', output: '/foo/' }); // or whatever the correct answer is, I'm not sure at this point
```

or

```javascript
testRelativizes({ base: 'http://example.org/////foo' }, relativize => {
  relativize('http://example.org/foo', /foo/');
  relativize('http://example.org/////foo', /foo/');
  // etc.
});
```

where `checkRelativizes`/`testRelativizes` instantiate a parser and make it serialize one triple, checking for this specific pattern. Similar to how the relative IRI parsing tests work (some good cases to be lifted from there).

But @smessie this might all be much more than what you bargained for, so don't feel obliged to take this on. It's just after seeing this solution that it becomes clear there's much, much more to it. And that was definitely one of the reasons why the `Writer` is conservative in abbreviating; it's hard to get it right for every single case, and possibly also computationally expensive.

src/N3Writer.js
```diff
@@ -58,9 +58,30 @@ export default class N3Writer {
...
+        const escapedSchemeDelimiter = escapeRegex('://');
```
RubenVerborgh, line 62: I can't think of an edge case right away, but I would do this in the raw domain rather than the escaped domain.

src/N3Writer.js
```diff
@@ -58,9 +58,30 @@ export default class N3Writer {
...
+        // We only want to make optional groups for the path fragments, not the host, so ignore the scheme delimiter ://
+        let postDelimiterIndex = escapedBaseIRI.indexOf(escapedSchemeDelimiter);
+        postDelimiterIndex = postDelimiterIndex >= 0 ? postDelimiterIndex + escapedSchemeDelimiter.length : 0;
+        // Make the path fragments optional by replacing (unescaped example) http://example.org/foo/bar/ with http://example.org/(foo(/bar/)?)?
+        let postDelimiter = escapedBaseIRI.substring(postDelimiterIndex);
+        while (postDelimiter.match(/\/[^()]+[^(]*$/)) {
```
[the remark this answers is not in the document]
smessie: Do you think of other problematic characters than `(`?
I think that character is problematic because I use it in my regex part to match previously added fragment groups.
So while we disallow `(` in the regex pattern, we can allow the escaped version `\(`, which would result in the following regex instead: `/\/[^()]+(\\\(|[^(])*$/`.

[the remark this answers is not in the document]
RubenVerborgh: It's still tricky to solve this in the escaped domain, an example being that backslashes can be escaped by backslashes, so the string `\\(` would cause trouble. While there are probably workarounds that count an uneven number of backslashes etc., this does not account for other possible vectors that are even harder to reason about.

It's inherently unnecessarily hard to search over regular expressions (or anything for that matter) in escaped form; the safest and simplest path forward is to have a single construction loop that creates the escaped expression, which is then only interpreted by a regular expression engine.

Regarding problematic characters in general, an option if to restrict ourselves to a subset of the IRI space over which stronger assumptions hold. That said, a single character-by-character (or segment-by-segment) loop over any input IRI should be able to safely and unambiguously transform it into a regular expression and corresponding parent prefix map, without requiring look-ahead.

Roughly:
1. The character string until the first slash is stored escaped but otherwise as-is, not counting any trailing slashes directly following the protocol colon towards the calculation of the first slash position. (If no such colon found, abort because of invalid absolute IRI.)
2. For any path segments and their position of occurrence, defined as `^/[^\/?#]*\/` starting from the character after each previous slash, construct a growing string `(?:(?:(?:…e1)e2)…en)` by starting from the empty string, prepending `(?:` and appending `en)` with `en` being the escaped path segment. Add the position to the lookup of parent prefixes. (Note, special behavior might be required for sequences of multiple slashes, dot slash, dot dot slash; or these could abort the parent path algorithm if unclear.)
3. Concatenate the strings from 1 an 2, which result in the required regex.

src/BaseIri.js
```diff
@@ -0,0 +1,132 @@
+export default class BaseIri {
+  constructor(iri) {
+    if (iri.startsWith('file://')) {
+      throw new Error('Base IRIs starting with file:// are not supported.');
+    }
```
RubenVerborgh, line 5: Maybe implicitly lacking support is okay. As in: don't throw, just don't act.

src/index.js
```diff
@@ -22,6 +23,7 @@ import {
 
   termFromId,
   termToId,
+// eslint-disable-next-line import/max-dependencies
```
RubenVerborgh, line 26: That's weird; probably should update the linter config then.

test/N3Writer-test.js
```diff
@@ -902,6 +903,4691 @@ describe('Writer', () => {
         writer.quadToString(new NamedNode('a'), new NamedNode('b'), new Quad(new NamedNode('a'), new NamedNode('b'), new NamedNode('c'), new NamedNode('g'))),
       ).toBe('<a> <b> <<<a> <b> <c> <g>>> .\n');
     });
+
+    testRelativizes('http://example.org/', [
```
RubenVerborgh, line 907: Cool cool cool. Thanks for the many, many tests.

How did you choose them?
Is there some kind of logic/structure to the test that comments could help capture?

src/BaseIri.js
```diff
@@ -0,0 +1,132 @@
...
+      baseIRIRegex += escapeRegex(iri.substring(0, match.index + match[0].length));
+      i = match.index + match[0].length;
+      stage = 1;
+    }
+    else {
+      throw new Error(`Base IRI should contain a scheme followed by its delimiter (e.g., http://). Instead, found: '${iri}'.`);
+    }
+
+    if (/\/\.{0,2}\//.test(iri.substring(i))) {
+      throw new Error('Base IRIs containing `//`, `/./`, or `/../` are not supported.');
```
RubenVerborgh, line 31: Perhaps also silent failure here is okay.

RubenVerborgh (commented):
Yeaaaah good stuff!!

test/N3Writer-test.js
smessie: For every baseIRI, the tested nodes to relativize are the same. If you have the list of nodes, each IRI is being used once as baseIRI. Then the list of tested nodes also contains their "extended" variant, so just the IRI appended with "extended" to test that things like `http://ex.org/foo` don't match `http://ex.org.fooextended` in the way that it would relativize to `<extended>`, but properly relativize to `<fooextended>`.
I can try to summarize this in a comment above the tests.

RubenVerborgh:
Thanks @smessie, looking really good. Great to have such extensive tests.
Did some minor cosmetic changes, and now merging this.
Will still do some refactorings on main, and then release a new vminor.

RubenVerborgh (approved):

PR #515, opened 2025-03-31
Remove queueMicrotask dependency

RubenVerborgh:
I think we're good: https://developer.mozilla.org/en-US/docs/Web/API/Window/queueMicrotask#browser_compatibility


jeswr (approved):
Never thought I'd see the day ... LGTM!

PR #529, opened 2025-05-26
feat: use origin based `</x/y/>` relativisation rather than `<../../../x/y/>` when result is a shorter relative IRI.

### LINKED ISSUE #528: feat: cleaner relative IRI writing
If I have 

```ttl
@base <http://example.org/a/b/c/d> .

</Jesse> a <#Thing> .
```

Then parsing and re-serialising this will currently result in the first term being serialised as `<../../../../Jesse>` rather than the cleaner `</Jesse>`.

jeswr:
Partially resolves #528 but does not cover the following test cases (where there are 2 or more sub paths after the relativisation):

```js
     relativizes('an IRI directly relative to the base with two sub paths', 'http://example.org/a/b/c/d/e/f/',
       'http://example.org/a/b/c/d/h/i/', '../../h/i/');

     relativizes('an IRI where it is better to use a / path [/x/y/z]', 'http://example.org/x/q/r/n/m/',
       'http://example.org/x/y/z', '/x/y/z');
```

## Summary by CodeRabbit

- **New Features**
  - Introduced an option to generate absolute IRIs starting with a slash for improved path resolution.
- **Bug Fixes**
  - Enhanced path replacement logic to conditionally use absolute or relative paths based on configuration and segment analysis.
- **Tests**
  - Added comprehensive tests validating both absolute and relative IRI generation behaviors under different settings.


RubenVerborgh:
Do we want this though? Because a big reason for relative support in the writer is the ability to move documents without changing syntax. CC @smessie

jeswr:
Right, that makes sense.

The context for opening this PR is that I use the `BaseIRI` class to relative IRIs in https://github.com/jeswr/pretty-turtle which has an explicit goal of maximal readability (or compactness depending on settings).

Given that I think this functionality would still be useful opt-in functionality in `pretty-turtle`. Would you be open to this being a separate opt-in functionality. If not I can just create a custom `BaseIRI` class downstream.

RubenVerborgh:
My two cents: pretty printing is a very different business from N3Writer. I'd architect the thing very differently for pretty printing (especially if the streaming constraint can be dropped).

It _could_ be an option on BaseIRI.js; wondering if it makes sense to forward base IRI options from N3Writer to BaseIRI. I'm mainly concerned about the slippery slope, in which N3Writer becomes no-actually-a-pretty-printer-but-some-features-are-getting-close-so-why-not, whereas the design goals are different. N3Writer is about writing Turtle fast (so any abbreviations are for speed more than readability), and features tend to serve a functional need (relative URIs are to facilitate moving). Pretty printers are inherently different beasts.

jeswr:
Withdrawing this together with #528 — see the close-out there. For the record, a correctness note anyone reviving this branch should know: the origin substitution currently assumes `segments[0]` marks the end of an authority, which only holds for `scheme://` bases. For authority-less bases that `BaseIRI.supports()` accepts, the flag produces wrong output: with base `urn:a/b/c/d/e/`, `toRelative('urn:a/x')` returns `/x`, which resolves back to `urn:/x` rather than `urn:a/x` (same failure for e.g. `s:/a/b/c/d/e/`; verified by parsing the writer output back with N3). The fix would be to only apply origin-based replacements in `_getBaseMatcher()` when the base actually has an authority (scheme match contains `//`), keeping the `../` form otherwise, plus regression tests for both. Two smaller points for posterity: the outputs are still relative references (path-absolute, RFC 3986 §4.2), so `absoluteIris` is a slightly misleading name — `originRelative` or similar would be clearer; and the branch would need a rebase now that RDF 1.2 has landed on main.

PR #531, opened 2025-05-30
Add N3Util.isQuad

opl-:
Will add DefinitelyTyped types when this is merged.


RubenVerborgh:
@jeswr No, not even a button for it. Something seems off. I rebased but no difference.

PR #586, opened 2026-05-05
RDF Messages support added

### LINKED ISSUE #439: Extending the writer with the possibility to write a comment to the outputStream
Recently we added support for optionally parsing comments. We should probably also have a public function in the writer that allows one to write a comment. I.e.:

```typescript
addComment(comment: string) {
   this._outputStream.push('# ' + comment);
}
```

In the StreamWriter, it could then be listening in on `comment` events from the input.

Does this sound reasonable? If so, I can open a PR.

pietercolpaert:
This is an implementation of the [RDF Messages specification](https://w3c-cg.github.io/rsp/spec/messages) by the W3C RDF Stream Processing CG that is now seeking for implementations of the spec.

More context:
 * This is also being discussed in the RDF JS data model spec: https://github.com/rdfjs/data-model-spec/issues/183
 * The poster paper we are presenting at the ESWC2026 conference: https://pietercolpaert.be/papers/eswc2026-rdf-messages/
  * @RubenVerborgh: we already discussed this in a slightly different context in the past as well, for which this now is I believe a better fix: https://github.com/rdfjs/N3.js/issues/439#issuecomment-2334155572

## Summary by CodeRabbit

* **New Features**
  * RDF message parsing support—enable via messages: true or version strings ending in -messages
  * Parser delivers grouped messages via an onMessage callback; stream parser emits message events
  * Writer can serialize message batches with addMessage(), using Turtle-style @message . delimiters

* **Behavior**
  * Blank-node labels scoped per message; quads still emitted immediately and also batched
  * MESSAGE / @message delimiters recognized; messages flushed at EOF

* **Tests**
  * Added coverage for parsing, streaming, serialization, delimiter rules, and edge cases

* **Documentation**
  * README documents message parsing, events, and serialization usage

### DIFF
--- a/README.md
+++ b/README.md
@@ -98,9 +98,11 @@ parser.parse(tomAndJerry,
   });
 ```
 The callback's first argument is an optional error value, the second is a quad.
+The fourth argument is an optional `messageCounter`, which is only defined when parsing in RDF message mode.
 If there are no more quads,
 the callback is invoked one last time with `null` for `quad`
 and a hash of prefixes as third argument.
+When parsing in RDF message mode, that final callback also receives the current `messageCounter`.
 <br>
 
 Alternatively, an object can be supplied, where `onQuad`, `onPrefix` and `onComment` are used to listen for `quads`, `prefixes` and `comments` as follows:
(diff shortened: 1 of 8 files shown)


pietercolpaert:
I believe there are no remaining issues here and this PR is ready to be checked by the maintainers.

RubenVerborgh:
I'd propose to only add features to N3.js main that are W3C recommendations.
Because all of this will be shipped in _any_ RDF parser, any browser version, etc.

Could this be implemented as a subclass of the Lexer and Parser, in a separate module?
If any blockers arise, we could add hooks for them in N3.js.

pietercolpaert:
> I'd propose to only add features to N3.js main that are W3C recommendations.

It’s our ambition that RDF Messages ends up in a W3C recommendation: either the core RDF one, either the RDF Stream Processing (RSP) working group that is being created. It’s however a chicken-egg situation: for it to move on we need implementations in important frameworks. The rule could be perceived as confusing: N3 itself is also not a recommendation, and RDF1.2 triple terms are also not yet a recommendation (of course the adoption in this library benefits the upcoming recommendation).

We are now promoting this work to the broader community now that the RDF Messages proposal has reached a decent draft state in the W3C RSP CG. Next week we’re promoting it at the ESWC conference.

> Because all of this will be shipped in any RDF parser, any browser version, etc.

There are 100 additional lines of code we propose here in the src folder. After building this for the browser, when I build `browser/n3.min.js` on my machine, the size remains exactly `276K` on the main and my PR branch.

Emotional argument: the RDF Messages proposal is worth championing as it should be close to your heart: it’s a basis for message stream logs that can provide a provenance trail for trustworthy data flows, where I see trust envelopes as a fundamental next piece of the puzzle.

PR #598, opened 2026-06-17
feat: configurable #match() semantics

### LINKED ISSUE #595: Configurable `#match()` semantics
Currently, the `DatasetCore` returned by `#match()` will not materialise until a mutation such as `#add()` is performed.

This has semantic implications. If the parent `Dataset` is mutated prior to a mutation on the child dataset, then all iterations over the child dataset, prior to the mutation will contain the mutation. The child dataset will receive all mutations on the parent dataset until such a time at which a mutation on the child dataset is performed.

We should enable configurations for the following semantics:
 - The current behavior (for backwards compatibility)
 - [I am confident we would want this] The `child` dataset is always based on the contents of the `Dataset` at the time of `#match()`. I would argue that this is the most correct interpretation of the RDF/JS specification. The question is whether this should become the default behavior of the library.
 - [I am confident we would want this] The `child` datasets mutations are "forwarded" to the `parent` dataset mutations. The child dataset _always_ reflects the state of the parent dataset (i.e. all parent mutations are also "fowarded" to the child).
 - [I am **not** confident we would want this] The `child` datasets maintain a diff of changes. They reflect the state of the parent dataset with the diffs incoroproated.

jeswr:
## What

Resolves #595: make the semantics of the dataset returned by `Store#match()` configurable via a `matchSemantics` option, both per call and as a store-wide default:

```js
store.match(s, p, o, g, { matchSemantics: 'snapshot' });
new Store([], { matchSemantics: 'forwarded' });
```

- **`lazy`** (default, unchanged) — current behaviour: the view delegates live to the parent until the first operation that materialises it (a mutation, or a materialising read such as `size` or `has`).
- **`snapshot`** — the view reflects the parent at `match()` time; later parent mutations never affect it.
- **`forwarded`** — the view always reflects the parent. Matching mutations are forwarded in, and pattern-constrained view mutations are written through. Nested forwarded matches stay connected to the root with all ancestor patterns applied.

The diff-based mode from #595 is intentionally not implemented (left as future work).

## Default and next major version

The default stays `lazy` in this PR. In the next major version, `snapshot` will become the default, and only mutating operations will materialize explicitly lazy views; reads such as `size` and `has` will leave them live.

## API decisions

- **`_detach()`** (internal method): freezes a view to its current contents and releases its parent observer. This deliberately materializes the view once and is not part of the public API. Observers hold views weakly: dropping a view allows it to be collected without explicit detachment. Cleanup records also reference the parent weakly.
- **Sub-views inherit their parent's semantics.** Supplying the same `matchSemantics` explicitly is allowed; a different valid value throws an error naming both modes, and an unknown value uses the normal validation error. Changing modes at a view boundary would make snapshot timing, materialisation, and the write target ambiguous; callers can create a separate view from the root when they need another mode.
- **Forwarding is recursive and pattern-constrained.** Nested forwarded views observe and write to the root using the intersection of every ancestor pattern. `add` and `addAll` stop and throw on a non-matching quad; as with the store's existing sequential `addAll`, earlier additions remain. `import` reports a stream error. `delete` throws for a quad outside the combined pattern, and `deleteMatches` throws for a conflicting deletion pattern. Compatible wildcard deletions remain scoped to the view, and deleting an absent matching quad is a no-op. Conflicting nested patterns produce a permanently empty view that rejects mutations.
- **Detachment is per view.** Detaching a parent does not silently detach descendants that were already returned to the caller.

## Notes (not visible from the diff)

- Non-`lazy` views attach a weak pre-mutation observer. A lazily initialized `FinalizationRegistry` removes collected registrations, with opportunistic pruning during mutation notification if finalization is delayed. An unmatched mutation only performs a pattern check; a matching mutation materialises or updates the view as needed. Materialized forwarded views update their shared numeric indexes directly without constructing intermediate quads. Duplicate additions return early on the observed path, before notification. A `forwarded` view that is never read is never materialised. Pattern term ids are cached after first resolution. A provably empty nested view does not attach an observer.
- Mutation *during* an in-progress iteration (sync iterator or async stream): every pass stays stable as of when it started. A matching mutation freezes a baseline for the active iteration generation, and open iterations switch to that baseline. Baselines are released when the last iteration closes (or the stream is destroyed).
- Store-level `matchSemantics` is validated at construction. Internal `match()` calls (`Store#deleteMatches`, `Store#toStream`, and view fallbacks) always use `lazy`, so a non-lazy store default cannot leak observers through them. An `''` graph pattern means the default graph, matching `getQuads`.
- The checked-in scenario benchmark is `perf/N3StoreMatchSemantics-perf.js`; comparative base/head measurements are below.

## Performance impact

The feature was benchmarked between exact revisions `a624b55` (base) and `a41e11e` (feature implementation) on an Apple M1 with 16 GB RAM, macOS/Darwin 25.6.0, and Node.js 22.21.1. Both revisions were built in isolated directories against the same dependencies.

The comparative harness used ABBA ordering (base, PR, PR, base) for each sample, forced GC between samples, and reported the median paired ratio. Two independent runs used a 128 × 128 store (16,384 quads) with:

- 16 full match/iteration passes (262,144 yielded quads);
- 16,384 view constructions;
- 65,536 parent mutations; and
- 128 open views for observer fan-out (about 8.4 million pattern checks per sample).

| Comparison | Relative cost |
| --- | ---: |
| PR default mutation path vs base, no observers | parity; no repeatable regression |
| PR default `lazy` view construction vs base | **1.09×** (+9%) in both runs |
| PR default `lazy` match + full iteration vs base | **1.02–1.04×** (+2–4%) |
| PR `snapshot` view construction vs PR `lazy` | **1.16–1.19×** (+16–19%) |
| PR `snapshot` match + full iteration vs PR `lazy` | **1.24–1.25×** (+24–25%) |
| PR `forwarded` match + full iteration vs PR `lazy` | **1.13–1.30×** (+13–30%) |
| PR observer fan-out vs PR with no observers | `snapshot`: **1.6–2.3×**; `forwarded`: **1.7–1.8×** |

The iterator-safety cleanup (`3b0db3a`) was also compared directly with `a41e11e` using the same paired method and a higher iteration workload. Median ratios were 0.98× for default iteration, 0.98× for snapshot iteration, 1.04× for forwarded iteration, 0.97× for default view construction, and 0.92× for observer fan-out. These differences were within run-to-run noise; no repeatable cleanup regression was found.

The final recursive-forwarding and pattern-safety change (`889f35d`) was compared directly with the preceding PR tip (`8fae640`) using the same ABBA method, with 9 measured rounds after 4 warm-ups on a 64 × 64 store. Default match iteration was 1.00×, forwarded iteration was 0.99×, and existing top-level observer fan-out was 0.99×. Matching `addAll` through a forwarded view measured 1.02× and 1.03× in two runs, reflecting the new validation but remaining within the harness's noise threshold. Recursive-child cases were compared with equivalent manually flattened root views at the preceding tip—rather than its behaviorally different snapshot children—and measured 1.05× for matching writes and 0.84× for mutation fan-out; the paired samples crossed parity in both cases, so neither is claimed as a repeatable regression or improvement.

The default-path cost is concentrated in constructing a view (validation and view state); normal parent mutations retain the allocation-free null guard. The non-default costs are opt-in: stable iteration adds generation bookkeeping, while mutation fan-out scales with the number of attached views. These figures are microbenchmarks and should be read as relative costs on this machine, not universal throughput numbers.

To run the checked-in scenario benchmark:

```sh
npm run build
node perf/N3StoreMatchSemantics-perf.js 128
```

## Review follow-up validation

Commit `846ecd8` addresses the remaining review feedback. The early return on `hasInIndex` is limited to the observed path, preserving the existing no-observer fast path. An isolated comparison with `f9b90ff` measured a median paired cost of **0.83×** for duplicate additions with an observer (range 0.76–0.96× across seven rounds), so the early return was retained.

The complete change was then compared with `f9b90ff` using identical esbuild Node bundles on an Apple M1, Darwin 27.0.0, Node.js 25.1.0. Each of seven rounds used fresh processes in ABBA order, four warm-up samples per scenario, and forced GC before timing. Each store held 8,192 quads; duplicate-add samples performed 131,072 calls, insertion/removal samples performed 8,192 calls, and the materialized forwarding sample performed 16,384 add/remove calls with 16 attached wildcard views. The other observed cases used one unmaterialized, non-matching forwarded view.

| Scenario | Median paired final/base cost | Paired range |
| --- | ---: | ---: |
| Duplicate additions, no observers | 1.00× | 0.94–1.07× |
| Duplicate additions, observer attached | **0.85×** | 0.82–0.89× |
| New additions, no observers | 1.00× | 0.92–1.11× |
| New additions, observer attached | 1.00× | 0.94–1.02× |
| Removals, no observers | 0.99× | 0.95–1.06× |
| Mutations with 16 materialized forwarded views | **0.86×** | 0.77–0.90× |

Both targeted improvements appeared in every paired round; ordinary mutation paths remained within run-to-run variation. These are machine-specific microbenchmarks. The checked-in scenario benchmark now includes duplicate additions and mutations with materialized views.

Validation passed:

- `npm test -- --runInBand --roots src test`: **6,991 tests, 100% coverage**. Explicit roots exclude pre-existing nested local worktrees.
- `npm run lint`
- `npm run build` (Node and both browser bundles)
- `node perf/N3StoreMatchSemantics-perf.js 32`

## Observer lifetime and memory regression coverage

Commit `c8abee0` addresses [the transient-view retention comment](https://github.com/rdfjs/N3.js/pull/598#discussion_r4064792598) for all callers. Store observer sets now contain `WeakRef`s to views instead of bound callbacks. Finalization records contain only weak references to the parent store and view; they cannot root either object graph. Explicit detachment unregisters the cleanup record, and mutation notification also prunes collected views if finalization has not run yet. Snapshot and forwarded behavior stays intact while a view or its active reader remains reachable. GC and finalizer timing remain runtime-controlled.

The normal Jest suite now runs isolated Node subprocesses with `--expose-gc`, loading the current source through Babel. For each observing mode, the memory tests reclaim 1,920 transient views in repeated batches and check that observer registrations, materialized backing stores, and iterator baselines are released. Additional scenarios cover nested views, discarded parents, `getRulesFromDataset`, and live iterators/streams across GC. Deterministic tests cover delayed finalization, stale cleanup requests, and dead-parent cleanup. The transient-view regression was verified to fail against the preceding implementation in both modes.

The observing modes require native `WeakRef` and `FinalizationRegistry` support ([Node.js 14.6+](https://v8.dev/features/weak-references#weakref-support), or browsers exposing both APIs). Unsupported runtimes get a clear error when selecting a non-lazy mode; importing N3 and using default lazy views still works without those APIs.

Validation for this follow-up: **7,009 tests pass with 100% coverage**, plus lint, Node and browser builds, and the scenario benchmark's correctness assertions. The earlier performance comparisons above describe their named revisions; they do not measure this weak-observer change.

Closes #595

### DIFF
--- a/README.md
+++ b/README.md
@@ -381,28 +381,29 @@ The store provides the following search methods
 ### Configuring `match()` semantics
 
 The dataset returned by `match()` is also a readable stream. By default, it is
-*lazy*: it delegates live to the parent store until its own first mutation,
-which means parent mutations made before that point are reflected in the view.
-You can choose a different behavior with the `matchSemantics` option, either
-per-call or as a store-wide default:
+*lazy*: it delegates live to the parent store until the first operation that
+materializes the view (a mutation, or a non-streaming read such as `size` or
+`has`), which means parent mutations made before that point are reflected in
+the view. You can choose a different behavior with the `matchSemantics` option,
+either as a store-wide default or per call:
 
 ```JavaScript
 import { Store, DataFactory } from 'n3';
 const { namedNode } = DataFactory;
 
-// Per call:
-const view = store.match(namedNode('s'), null, null, null, { matchSemantics: 'snapshot' });
-
-// Or as the default for every match() of a store:
+// As the default for every match() of a store:
 const store = new Store([], { matchSemantics: 'snapshot' });
+
+// Or per call:
+const view = store.match(namedNode('s'), null, null, null, { matchSemantics: 'snapshot' });
 ```
 
 Supported values:
 
 - `'lazy'` (default) — backwards-compatible behavior. The view reflects the
-  parent store until the view itself is first mutated, after which it is frozen
-  to a snapshot. Parent mutations made before that first mutation leak into the
-  view.
+  parent store until the first operation that materializes it (a mutation, or a
+  non-streaming read such as `size` or `has`), after which it is frozen to a
+  snapshot. Parent mutations made before that point leak into the view.
 - `'snapshot'` — the view reflects the parent contents *at the time of*
   `match()`. Later parent mutations never affect it. This is the most
   spec-correct interpretation of an RDF/JS dataset.
--- a/README.md
+++ b/README.md
@@ -409,6 +409,9 @@ Supported values:
 - `'forwarded'` — the view always reflects the parent state: matching parent
   mutations are forwarded to the view, and mutations on the view (`add`,
   `delete`, `addAll`, `deleteMatches`) are written through to the parent.
+  View mutations act on the parent unrestricted by the view's pattern: adding
+  a quad that does not match the pattern mutates the parent but never appears
+  in the view, and `deleteMatches` can delete parent quads outside the view.
   Calling `match()` on a `'forwarded'` view returns a `'snapshot'` sub-view:
   nested views never write through to the root store.
 
--- a/README.md
+++ b/README.md
@@ -378,6 +378,44 @@ The store provides the following search methods
 - `getGraphs` returns an array of unique graphs occurring in matching quad
 - `forGraphs` executes a callback on unique graphs occurring in matching quads
 
+### Configuring `match()` semantics
+
+The dataset returned by `match()` is also a readable stream. By default it is
+*lazy*: it delegates live to the parent store until its first own mutation,
+which means parent mutations made before that point are reflected in the view.
+You can choose a different behavior with the `matchSemantics` option, either
+per call or as a store-wide default:
+
+```JavaScript
+import { Store, DataFactory } from 'n3';
+const { namedNode } = DataFactory;
+
+// Per call:
+const view = store.match(namedNode('s'), null, null, null, { matchSemantics: 'snapshot' });
+
+// Or as the default for every match() of a store:
+const store = new Store([], { matchSemantics: 'snapshot' });
+```
+
+Supported values:
+
+- `'lazy'` (default) — backwards-compatible behavior. The view reflects the
+  parent store until the view itself is first mutated, after which it is frozen
+  to a snapshot. Parent mutations made before that first mutation leak into the
+  view.
+- `'snapshot'` — the view reflects the parent contents *at the time of*
+  `match()`. Later parent mutations never affect it. This is the most
+  spec-correct interpretation of an RDF/JS dataset.
+- `'forwarded'` — the view always reflects the parent state: matching parent
+  mutations are forwarded to the view, and mutations on the view (`add`,
+  `delete`, `addAll`, `deleteMatches`) are written through to the parent.
+
+For `'snapshot'` and `'forwarded'`, an iteration (synchronous or via the
+stream) that is already in progress keeps a stable view of the quads as of when
+it started, even if a matching parent mutation lands mid-iteration. Views are
+maintained incrementally — they only do work when a parent mutation actually
+matches the view's pattern — so the common path stays fast.
+
 ## Reasoning
 
 N3.js supports reasoning as follows:
--- a/README.md
+++ b/README.md
@@ -418,6 +418,14 @@ it started, even if a matching parent mutation lands mid-iteration. Views are
 maintained incrementally — they only do work when a parent mutation actually
 matches the view's pattern — so the common path stays fast.
 
+Such views observe the parent store: a `'snapshot'` view until its contents are
+first needed (its first matching parent mutation or first materializing read),
+a `'forwarded'` view for the lifetime of the store. The store holds a strong
+reference to each observing view — retaining it against garbage collection —
+and checks every mutation against each open view's pattern. Call
+`view.detach()` to release a view early: it is frozen to its contents at detach
+time and no longer taxes the store.
+
 ## Reasoning
 
 N3.js supports reasoning as follows:
--- a/README.md
+++ b/README.md
@@ -409,7 +409,8 @@ Supported values:
   spec-correct interpretation of an RDF/JS dataset.
 - `'forwarded'` — the view always reflects the parent state: matching parent
   mutations are forwarded to the view, and mutations on the view (`add`,
-  `delete`, `addAll`, `deleteMatches`) are written through to the parent.
+  `delete`, `addAll`, `deleteMatches`, `import`) are written through to the
+  parent.
   View mutations act on the parent unrestricted by the view's pattern: adding
   a quad that does not match the pattern mutates the parent but never appears
   in the view, and `deleteMatches` can delete parent quads outside the view.
--- a/README.md
+++ b/README.md
@@ -417,6 +417,9 @@ Supported values:
   Calling `match()` on a `'forwarded'` view returns a `'snapshot'` sub-view:
   nested views never write through to the root store.
 
+`match()` on a view takes no `matchSemantics` option: a sub-view inherits the
+view's semantics (with `'forwarded'` downgraded to `'snapshot'` as above).
+
 For `'snapshot'` and `'forwarded'`, an iteration (synchronous or via the
 stream) that is already in progress keeps a stable view of the quads as of when
 it started, even if a matching parent mutation lands mid-iteration. Views are
--- a/README.md
+++ b/README.md
@@ -409,6 +409,8 @@ Supported values:
 - `'forwarded'` — the view always reflects the parent state: matching parent
   mutations are forwarded to the view, and mutations on the view (`add`,
   `delete`, `addAll`, `deleteMatches`) are written through to the parent.
+  Calling `match()` on a `'forwarded'` view returns a `'snapshot'` sub-view:
+  nested views never write through to the root store.
 
 For `'snapshot'` and `'forwarded'`, an iteration (synchronous or via the
 stream) that is already in progress keeps a stable view of the quads as of when
--- a/README.md
+++ b/README.md
@@ -380,11 +380,11 @@ The store provides the following search methods
 
 ### Configuring `match()` semantics
 
-The dataset returned by `match()` is also a readable stream. By default it is
-*lazy*: it delegates live to the parent store until its first own mutation,
+The dataset returned by `match()` is also a readable stream. By default, it is
+*lazy*: it delegates live to the parent store until its own first mutation,
 which means parent mutations made before that point are reflected in the view.
 You can choose a different behavior with the `matchSemantics` option, either
-per call or as a store-wide default:
+per-call or as a store-wide default:
 
 ```JavaScript
 import { Store, DataFactory } from 'n3';
--- a/README.md
+++ b/README.md
@@ -503,6 +503,9 @@ Supported values:
   Calling `match()` on a `'forwarded'` view returns a `'snapshot'` sub-view:
   nested views never write through to the root store.
 
+`match()` on a view takes no `matchSemantics` option: a sub-view inherits the
+view's semantics (with `'forwarded'` downgraded to `'snapshot'` as above).
+
 For `'snapshot'` and `'forwarded'`, an iteration (synchronous or via the
 stream) that is already in progress keeps a stable view of the quads as of when
 it started, even if a matching parent mutation lands mid-iteration. Views are
--- a/README.md
+++ b/README.md
@@ -468,7 +468,7 @@ The store provides the following search methods
 
 The dataset returned by `match()` is also a readable stream. By default, it is
 *lazy*: it delegates live to the parent store until the first operation that
-materializes the view (a mutation, or a non-streaming read such as `size` or
+materializes the view (a mutation, or a materializing read such as `size` or
 `has`), which means parent mutations made before that point are reflected in
 the view. You can choose a different behavior with the `matchSemantics` option,
 either as a store-wide default or per call:
@@ -486,8 +486,8 @@ Supported values:
 
 - `'lazy'` (default) — backwards-compatible behavior. The view reflects the
   parent store until the first operation that materializes it (a mutation, or a
-  non-streaming read such as `size` or `has`), after which it is frozen to a
-  snapshot. Parent mutations made before that point leak into the view.
+  materializing read such as `size` or `has`), after which it is frozen to a
+  snapshot. Parent mutations made before that point remain visible in the view.
 - `'snapshot'` — the view reflects the parent contents *at the time of*
   `match()`. Later parent mutations never affect it. This is the most
   spec-correct interpretation of an RDF/JS dataset.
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -378,6 +378,44 @@ The store provides the following search methods
 - `getGraphs` returns an array of unique graphs occurring in matching quad
 - `forGraphs` executes a callback on unique graphs occurring in matching quads
 
+### Configuring `match()` semantics
+
+The dataset returned by `match()` is also a readable stream. By default it is
+*lazy*: it delegates live to the parent store until its first own mutation,
+which means parent mutations made before that point are reflected in the view.
+You can choose a different behavior with the `matchSemantics` option, either
+per call or as a store-wide default:
```
TallTed:
```suggestion
The dataset returned by `match()` is also a readable stream. By default, it is
*lazy*: it delegates live to the parent store until its own first mutation,
which means parent mutations made before that point are reflected in the view.
You can choose a different behavior with the `matchSemantics` option, either
per-call or as a store-wide default:
```

src/N3Store.js
```diff
@@ -328,6 +337,33 @@ export default class N3Store {
...
+    // Restore the allocation-free fast path once no observers remain.
+    if (this._observers.size === 0)
+      this._observers = null;
+  }
+
+  // ### `_notifyObservers` invokes all registered observers with the numeric
+  // ids of a quad that is about to be added or removed.
+  _notifyObservers(subjectId, predicateId, objectId, graphId, added) {
+    // Iterate over a copy so observers may unregister themselves while reacting.
+    for (const observer of [...this._observers])
```
jeswr: Why is this converted to an array for iteration. Is this not a performance issue?

README.md
jeswr: Thanks @TallTed — applied in 4bd5c5f with your phrasing, plus one accuracy fix in 751b952: a lazy view materializes not only on its own first mutation but on any non-streaming read (`size`, `has`, …), so the text now names that full trigger set.

src/N3Store.js
jeswr: The copy existed so an observer could unregister itself mid-notification (a snapshot view detaches when it materializes). Since deleting the current element during `Set` iteration is spec-safe, the copy was unnecessary and is removed in abbd107 — `_notifyObservers` now iterates the live set, and 86aeb1c adds a test pinning two observers self-detaching in one notification pass. Note this never touched the default path: `_notifyObservers` is only called when `this._observers !== null`, i.e. only while a snapshot/forwarded view is observing. Measured on that observer path (256 live forwarded views, 65k non-matching mutations, 7 reps): live-set iteration min 471ms / median 707ms vs the old copy min 707ms / median 1879ms — the per-mutation array allocation mostly showed up as GC pressure.

src/N3Store.js
```diff
@@ -1189,6 +1362,10 @@ class DatasetCoreAndReadableStream extends Readable {
   }
 
   deleteMatches(subject, predicate, object, graph) {
+    if (this._semantics === 'forwarded') {
+      this.n3Store.deleteMatches(subject, predicate, object, graph);
+      return this;
+    }
```
jeswr, line 1495: There is quite a lot of logic similar to this, lets have an internal getter to get the store to modify that they all use

jeswr: Addressed in 5679649: added a `_mutationStore` getter and routed `add`, `delete`, `addAll`, `deleteMatches`, and `import` through it. Forwarded return values and the low-level add/remove path remain unchanged. Build, lint, and all 6,926 tests pass.

src/N3Store.js
```diff
@@ -1347,23 +1500,63 @@ class DatasetCoreAndReadableStream extends Readable {
...
+  *_iterateStable() {
+    // Use a snapshot if the parent changes mid-iteration
+    const generation = this._generation;
+    let yielded = 0;
+    this._activeIterators++;
+    this._currentIterators++;
+    try {
+      const source = this._filtered ?
+        this._filtered[Symbol.iterator]() :
+        this.n3Store.readQuads(this.subject, this.predicate, this.object, this.graph);
```
jeswr: duplicate of above logic for `Symbol.iterator`

jeswr: Resolved in 3b0db3a by extracting `_sourceIterator()` and using it from both `[Symbol.iterator]()` and `_iterateStable()`. The helper stays inside the generators, so source selection still occurs on first read; regression tests cover lazy, snapshot, and forwarded timing.

src/N3Store.js
```diff
@@ -453,6 +480,10 @@ export default class N3Store {
     predicate = this._termToNewNumericId(predicate);
     object    = this._termToNewNumericId(object);
 
+    // Notify observers only for new quads
+    if (this._observers !== null && !hasInIndex(graphItem.subjects, subject, predicate, object))
```
jeswr, line 482: Rather than checking `!hasInIndex` cant we just move this to line 460 as that code path is only reached when the quad is not already in the index.

src/N3Store.js
```diff
@@ -518,6 +549,10 @@ export default class N3Store {
         !(object in predicates))
       return false;
 
+    // Notify observers before the mutation
+    if (this._observers !== null)
```
jeswr, line 551: If there is no performance penalty, please remove all instances of this check inside the `_notifyObservers` call.

src/N3Store.js
jeswr: This notification has to occur before the first index insertion: a snapshot observer materializes the pre-mutation store in `_onParentMutation`, so moving the call below `_addToIndex` would allow the newly added quad into the snapshot. The existing `should snapshot an absent pattern term before its first matching mutation` regression test exercises exactly that ordering. I also benchmarked insertion-hook versions that combined the existence check with pre-insert notification, but they regressed the duplicate-add hot path. I therefore retained the guarded precheck and clarified its timing requirement in 64b9ac5.

src/N3Store.js
jeswr: I benchmarked this before changing it. Moving the null guard into `_notifyObservers` makes the no-observer add path run `hasInIndex` unconditionally; the median duplicate-add cost increased from 578.0 to 635.4 ns/op (+9.9%). The no-observer removal case also increased from 1264.7 to 1288.0 ns/op (+1.8%). Focused insertion-hook variants that avoided the extra lookup still showed a double-digit duplicate-add regression, so I have retained the allocation-free call-site guards as the performance-positive option.

src/N3Store.js
```diff
@@ -1203,20 +1239,148 @@ function indexMatch(index, ids, depth = 0) {
...
  */
 class DatasetCoreAndReadableStream extends Readable {
   constructor(n3Store, subject, predicate, object, graph, options) {
     super({ objectMode: true });
     Object.assign(this, { n3Store, subject, predicate, object, graph, options });
+    const semantics = this._semantics = validateMatchSemantics(options.matchSemantics);
+
+    if (options.matchesNothing) {
+      this._matchesNothing = true;
+      this._filtered = new N3Store({ factory: n3Store._factory, entityIndex: options.entityIndex });
```
jeswr, line 1274: If we have forwarding semantics in place and a quad matching the pattern is added to the parent dataset - will it be appropriately forwarded?

src/N3Store.js
```diff
@@ -456,6 +478,10 @@ export default class N3Store {
     predicate = this._termToNewNumericId(predicate);
     object    = this._termToNewNumericId(object);
 
+    // Notify observers before inserting a new quad so snapshots retain their prior contents
+    if (this._observers !== null && !hasInIndex(graphItem.subjects, subject, predicate, object))
```
jeswr, line 482: Should assess performance impact of doing an early return when `hasInIndex`

README.md
```diff
@@ -464,6 +464,63 @@ The store provides the following search methods
...
+const store = new Store([], { matchSemantics: 'snapshot' });
+const snapshot = store.match(namedNode('s'));
+const forwarded = store.match(namedNode('s'), null, null, null, { matchSemantics: 'forwarded' });
+```
+
+Supported values:
+
+- `'lazy'` (default) — backwards-compatible behavior. The view reflects the
+  parent store until the first operation that materializes it (a mutation, or a
+  materializing read such as `size` or `has`), after which it is frozen to a
```
jeswr, line 489: In the next mver we should only make mutating operations materialisation operations.

README.md
```diff
@@ -464,6 +464,63 @@ The store provides the following search methods
...
+  snapshot. Parent mutations made before that point remain visible in the view.
+- `'snapshot'` — the view reflects the parent contents *at the time of*
+  `match()`. Later parent mutations never affect it. This is the most
+  spec-correct interpretation of an RDF/JS dataset.
```
jeswr, line 493: We should document that this will become the default in the next mver.

README.md
```diff
@@ -464,6 +464,63 @@ The store provides the following search methods
...
+- `'forwarded'` — the view always reflects the parent state: matching parent
+  mutations are forwarded to the view, and mutations on the view (`add`,
+  `delete`, `addAll`, `deleteMatches`, `import`) are written through to the
+  parent. Nested matches remain forwarded to the root, with every ancestor's
+  pattern applied. `add` and `addAll` throw upon encountering a quad outside
+  that combined pattern, while `import` emits an error on its input stream.
+  `delete` ignores a quad outside the pattern, and `deleteMatches` only removes
+  matching quads visible in the view.
```
jeswr, line 501: For consistency we should error here as well on pattern mismatch.

src/N3Store.js
```diff
@@ -1273,7 +1437,45 @@ class DatasetCoreAndReadableStream extends Readable {
...
+    this._detachObserver();
+    // Lazy views deferred this state
+    if (this._semantics === 'lazy') {
+      this._generation = 0;
+      this._activeIterators = this._currentIterators = 0;
+      this._baselines = null;
+    }
+    this._semantics = 'snapshot';
+    return this;
+  }
```
jeswr, line 1469: This should be an internal method and marked with an underscore.

src/N3Store.js
jeswr: Oh right - nothing will ever match, because this comes from a pattern where there is no possible match.

src/N3Store.js
```diff
@@ -1203,20 +1239,148 @@ function indexMatch(index, ids, depth = 0) {
...
+    this._freezeCurrentIterators();
+
+    // Keep using the parent until materialization
+    if (this._semantics === 'forwarded') {
+      if (this._filtered) {
+        const quad = this._toQuad(subjectId, predicateId, objectId, graphId);
+        if (added)
+          this._filtered.addQuad(quad);
+        else
+          this._filtered.removeQuad(quad);
```
jeswr, line 1354: We shouldn't need to convert to the intermediate quad representation using `_toQuad` because `_filtered` should be using the same entity index. Therefore we should be able to work directly with the index values here.

PR #602, opened 2026-06-18
fix: StreamParser failure on language tags split across chunks

sarathfrancis90:
Fixes #578.

## Bug
`StreamParser` throws `Unexpected "-nl"` when a multi-subtag language tag such as `@nl-nl` happens to be split across two chunks. The same input parsed in one piece with `Parser.parse()` works fine.

## Cause
The lexer's language-code regex (`_langcode`) has a trailing lookahead `(?=[^a-z0-9])`, which a dash satisfies. So when the buffer ends in `@nl-` (or `@nl-nl` with nothing after it yet), the regex matches only the first subtag `nl`. With streaming input the lexer then emits `langcode "nl"` and chokes on the remaining `-nl` in the next chunk.

## Fix
In the `@` branch, if the input is not yet finished and the language-code match is immediately followed by a dash, defer the match so the remaining subtags can arrive in a later chunk. This mirrors the existing chunk-boundary handling already used for blank nodes, numbers and prefixed names.

## Testing
Added lexer tests (subtag split, and split exactly on a subtag boundary) and a `StreamParser` test, each reproducing the error before the fix. I also split the literal `"x"@nl-nl-foo .` at every position and confirmed all of them now parse to `nl-nl-foo`. Full suite passes (6496 tests) with 100% coverage, and eslint is clean.

## Release Notes

* **Bug Fixes**
  * Fixed tokenization of language tags with subtags when parsing streamed or incomplete input, preventing premature token emission and ensuring correct subtag recognition across chunk boundaries.

* **Tests**
  * Added test cases for language tag parsing when subtag boundaries span across multiple stream chunks.


jeswr:
Thanks @sarathfrancis90 — spot-on diagnosis of #578: the `(?=[^a-z0-9])` lookahead is satisfied by the dash itself, so a chunk ending in `@nl-` prematurely emits `nl`, and deferring the match until more input arrives is exactly the right idea.

The deferral as written also triggered when the character after the language tag is the first dash of a direction code (`--ltr`/`--rtl`). Since the lexer treats "no token found + newline in the buffer" as a syntax error, valid streamed input like `'"x"@en--rtl .\n"y"@en .'` (with the chunk still unfinished) threw `Unexpected "@en--rtl"` on this branch while parsing fine on `main`.

Per @jeswr I've pushed directly to this branch: a one-line refinement that defers only when the dash could still extend the language tag (i.e. not when it is followed by another dash), plus lexer tests for the directional cases (`@en--rtl` with a buffered newline, `@en-` / `-rtl` split) and the remaining `@nl-nl` split point (`@nl-n` / `l`). Your original fix and tests are untouched; `main` was merged in for the updated toolchain. Lint and full tests pass at 100% coverage.

Ready for @jeswr's re-review.

TallTed:
@jeswr — You seem to be [talking to yourself](https://github.com/rdfjs/N3.js/pull/602#issuecomment-4870191168). "Per @jeswr I've pushed" posted *by* @jeswr is quite confusing. I'm guessing that part of what's happening is _AI-ish_. I strongly advise that only *you* post as you, and that you do not talk to yourself nor approve yourself nor advise yourself.

jeswr:
>  I'm guessing that part of what's happening is AI-ish. I strongly advise that only you post as you, and that you do not talk to yourself nor approve yourself nor advise yourself.

You are right, and agreed that I need to improve the distinction here. In the first instance I have tagged all new PR's which have been AI generated. I am also working on setting up a bot account for all AI generated content - and formalising a policy for best practises for handling AI generated content.

jeswr (approved):

PR #612, opened 2026-07-02
chore(ci): cache rdf-test-suite fixtures in spec jobs

### LINKED ISSUE #609: spec test flakiness
Observed in https://github.com/rdfjs/N3.js/actions/runs/28062380806/job/83079374825 - this is the second time I have seen such an error within 24 hours.

jeswr:
Spec flakes (`Expected: []`, ENOENT on `.rdf-test-suite-cache/*.headers`) come from a race in rdf-test-suite's fetch cache: `Util.fetchCached` writes non-atomically while test cases are constructed with `Promise.all`, and many tests share the same `mf:result` file — a concurrent fetch of the same URL sees `existsSync` pass and reads a truncated body. All 40 spec jobs currently start cold, maximising exposure. This persists `.rdf-test-suite-cache/` with `actions/cache`; warm runs are served entirely from cache with no fetches or writes, closing the race window.

Notes on the key:
- Per suite, shared across Node versions (identical fetch set, ~1 MB per entry).
- Cache hits are served without revalidation, so fixtures are pinned once saved; refreshing them (e.g. rdf12 manifest updates) is a deliberate `v1` bump rather than silent churn on unrelated PRs.
- `hashFiles('package.json')` rotates the key when the spec scripts (manifest URLs, `--skip` lists) may have changed — an exact hit skips the save, so without rotation any new URLs would be refetched live on every run forever — and `restore-keys` seeds the rotated cache so only genuinely new files are fetched.

Verified locally that a cold `spec-1-1-nquads` run writes the cache at the repo root and a warm rerun passes. The real fix belongs upstream (atomic cache writes / in-flight fetch dedupe in `Util.fetchCached`); this keeps our CI robust in the meantime.

Closes #609

cc @jeswr


.github/workflows/ci.yml
```diff
@@ -66,6 +66,19 @@ jobs:
...
+      # served without revalidation, so entries are pinned once saved — bump v1
+      # to refetch the suites. hashFiles(package.json) rotates the key whenever
+      # the spec scripts may have changed; restore-keys seeds the rotated cache
+      # so only new files are fetched live.
+      - uses: actions/cache@v6
+        with:
+          path: .rdf-test-suite-cache
+          key: rdf-test-suite-cache-v1-${{ matrix.spec }}-${{ hashFiles('package.json') }}
+          restore-keys: |
+            rdf-test-suite-cache-v1-${{ matrix.spec }}-
```
jeswr: Since the manifests are served from the web, it is possible that they change from time to time. Especially the RDF 1.2 ones as the WG is still in flight. Can we have a cron job which looks for valid changes in the test suite - say weekly, or by manual trigger - and does something to trigger an update in the cache (like opening a PR to change the version number here).

jeswr: Added in f3759cd. A weekly (or manually dispatched) `refresh-spec-cache` workflow queries the GitHub API for the latest `w3c/rdf-tests@gh-pages` commit touching `rdf/rdf11` and `rdf/rdf12` (gh-pages is the branch Pages actually serves), writes it to a new `spec/cache-key.txt` pin, and opens a PR via `peter-evans/create-pull-request@v8` only when the pin changed. The pin is hashed into this cache key **and** the restore-keys prefix: since rdf-test-suite serves cache hits without revalidation, seeding a rotated key from the old cache would shadow the updated fixtures, so a pin bump deliberately starts fully cold (a `package.json`-only change still warm-seeds, as it cannot change what upstream serves).

I first tried the more direct approach of digesting a freshly fetched `.rdf-test-suite-cache`, but two fresh full fetches produced different file *sets* (rdf-test-suite's cache writes are fire-and-forget — the same non-atomicity as #609), so that digest can never be stable; the upstream commit SHA is deterministic instead. Verified locally: both workflows parse, the pin step reproduces the committed pin byte-for-byte (idempotent, so no-change weeks open no PR), and a malformed API response fails the step rather than writing a bogus pin. Not verifiable locally: the create-pull-request action end-to-end, and note that CI won't auto-run on the PRs it opens (`GITHUB_TOKEN` limitation) — they need a close/reopen or push to trigger checks.

.github/workflows/ci.yml
```diff
@@ -66,6 +66,23 @@ jobs:
...
+      # cache writes are not atomic, so on a cold cache concurrent fetches of
+      # shared fixtures can race and read truncated files (#609). Hits are
+      # served without revalidation, so entries are pinned once saved.
+      # spec/cache-key.txt pins the upstream w3c/rdf-tests revision; the
+      # refresh-spec-cache workflow PRs a bump when upstream changes. The pin
+      # is part of the restore-keys prefix too: stale entries are never
+      # revalidated, so a pin bump must start fully cold instead of seeding
+      # from an old cache that would shadow the updated fixtures. A
+      # package.json change still seeds from the same pin, as it cannot
+      # change what upstream serves.
```
jeswr:
```suggestion
```

.github/workflows/refresh-spec-cache.yml
```diff
@@ -0,0 +1,47 @@
+# The spec jobs in ci.yml cache the rdf-test-suite fixtures, and cache hits
+# are served without revalidation, so a saved cache never sees upstream
+# changes. This workflow watches the branch that actually serves the
+# manifests (w3c/rdf-tests@gh-pages, published at w3c.github.io/rdf-tests)
+# and, when the rdf11/rdf12 suites change, opens a PR that bumps
+# spec/cache-key.txt. Merging that PR rotates the CI cache key, so the spec
+# jobs refetch the suites cold (deliberately without seeding from the old
+# cache, which would shadow the updated fixtures).
```
jeswr:
```suggestion
```

PR #613, opened 2026-07-02
fix: parse literal subjects and predicates in N3 mode

### LINKED ISSUE #328: Literals should be valid subjects in text/n3
See https://github.com/eyereasoner/eye/issues/76#issuecomment-1450996748. Currently an error is thrown when they are encountered.

jeswr:
In `text/n3` mode, `_completeSubjectLiteral` dropped the look-ahead token when it was not a datatype or language tag, so a plain string literal as subject (`"1" <a> <b>.`) swallowed the predicate and failed one token later with `Expected entity but got .`. Literals in predicate position were rejected outright, although the N3 grammar allows a literal as a path item in both positions.

`_completeSubjectLiteral` now consumes the look-ahead token (as `_completeObjectLiteralPost` already did) and gains the same `!completed` guard as its object counterpart — previously an error while completing a subject literal (e.g. an undefined datatype prefix) crashed with a `TypeError` in callback mode instead of reporting the parse error. `_readPredicate` accepts literals (incl. datatype, language, and direction) in N3 mode only.

Turtle/TriG still reject literal subjects and predicates (negative tests added); the predicate error message changes from `Expected entity but got literal` to `Unexpected literal`, matching subject position. Tests include the six cases from the issue thread plus the formula cases.

Closes #328

cc @jeswr

### DIFF
--- a/src/N3Parser.js
+++ b/src/N3Parser.js
@@ -302,6 +302,18 @@ export default class N3Parser {
       // Additional semicolons can be safely ignored
       return this._predicate !== null ? this._readPredicate :
              this._error('Expected predicate but got ;', token);
+    case 'literal':
+      if (!this._n3Mode)
+        return this._error('Unexpected literal', token);
+
+      if (token.prefix.length === 0) {
+        this._literalValue = token.value;
+        return this._completePredicateLiteral;
+      }
+      else
+        this._predicate = this._factory.literal(token.value, this._factory.namedNode(token.prefix));
+
+      break;
     case '[':
       if (this._n3Mode) {
         // Start a new quad with a new blank node as subject
@@ -603,6 +615,8 @@ export default class N3Parser {
       const term = this._factory.literal(this._literalValue, { language: this._literalLanguage, direction: token.value });
       if (component === 'subject')
         this._subject = term;
+      else if (component === 'predicate')
+        this._predicate = term;
       else
         this._object = term;
       this._literalLanguage = undefined;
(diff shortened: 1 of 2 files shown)


src/N3Parser.js
```diff
@@ -302,6 +302,18 @@ export default class N3Parser {
...
+    case 'literal':
+      if (!this._n3Mode)
+        return this._error('Unexpected literal', token);
+
+      if (token.prefix.length === 0) {
+        this._literalValue = token.value;
+        return this._completePredicateLiteral;
+      }
+      else
+        this._predicate = this._factory.literal(token.value, this._factory.namedNode(token.prefix));
```
jeswr, line 314: Is this the same logic that exists for parsing objects already. If so can it be refactored.

jeswr (commented):
LGTM. Patch change.

src/N3Parser.js
jeswr: Refactored in cf581a2. The datatype/language tail was already shared with the object path via `_completeLiteral`/`_readDirCode`; what this PR duplicated was the completion wrapper, where `_completePredicateLiteral` cloned `_completeSubjectLiteral` (only the term slot, `_validAnnotation`, and continuation differed). Both now route through a shared `_completeTermLiteral(token, component)`. I left `_completeObjectLiteral` separate: its continuation goes through `_completeObjectLiteralPost` (list-item emission + `_getContextEndReader`) rather than advancing to the next statement component, so folding it in would add a three-way branch for no net dedupe.

jeswr, line 314: Is this generic logic that can be applied from the subject case and refactored

PR #614, opened 2026-07-02
fix: use prefixes for local names with interior dots

### LINKED ISSUE #372: N3 writer doesn't prefix integers correctly
`const N3 = require('n3');
const { DataFactory } = N3;
const { namedNode, literal, defaultGraph, quad } = DataFactory;
const writer = new N3.Writer({ prefixes: { ex1: 'http://example.org/' , ex2: 'http://example.org/_' } }); 
writer.addQuad(
    namedNode('http://example.org/s1'),                 
    namedNode('http://example.org/p'), 
    namedNode('http://example.org/1')             
);
writer.addQuad(quad(
    namedNode('http://example.org/s2'),  
    namedNode('http://example.org/p'),  
    namedNode('http://example.org/_1')                       
));
writer.end((error, result) => console.log(result));`

expected results:
`@prefix ex1: <http://example.org/>.
@prefix ex2: <http://example.org/_>.

ex1:s1 ex1:p ex1:1>.
ex1:s2 ex1:p ex2:1.`

results:
`@prefix ex1: <http://example.org/>.
@prefix ex2: <http://example.org/_>.

ex1:s1 ex1:p <http://example.org/1>.
ex1:s2 ex1:p ex1:_1.
`

jeswr:
The writer's prefix matcher rejected any dot in the local part, so IRIs like `http://a.org/v1.0` were always written in full even with a matching prefix. Turtle's `PN_LOCAL` allows interior dots, so the matcher now accepts them (`a:v1.0`, `a:a.b.c`) while local names with leading or trailing dots still fall back to full IRIs. Each step of the new pattern consumes at least one non-dot character, so there is no backtracking blow-up (checked against pathological inputs).

Two deliberate limits: consecutive interior dots (`a..b`) are legal `PN_LOCAL` but keep the full-IRI fallback to keep the pattern simple — the fallback is always correct output. And the slash question raised in the thread is out of scope: a raw `/` is not valid in a prefixed local name (that is `PN_LOCAL_ESC` territory) and extending the matcher that way carries the regex-performance risk already noted there.

Closes #372

cc @jeswr

### DIFF
--- a/src/N3Writer.js
+++ b/src/N3Writer.js
@@ -299,7 +299,7 @@ export default class N3Writer {
       }
       IRIlist = escapeRegex(IRIlist, /[\]\/\(\)\*\+\?\.\\\$]/g, '\\$&');
       this._prefixRegex = new RegExp(`^(?:${prefixList})[^\/]*$|` +
-                                     `^(${IRIlist})([_a-zA-Z0-9][\\-_a-zA-Z0-9]*)$`);
+                                     `^(${IRIlist})([_a-zA-Z0-9](?:\\.?[\\-_a-zA-Z0-9])*)$`);
     }
     // End a prefix block with a newline
     this._write(hasPrefixes ? '\n' : '', done);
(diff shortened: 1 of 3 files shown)


jeswr (commented):
LGTM. Patch fix.

PR #615, opened 2026-07-02
fix: add opt-in formula-only blank node scoping in N3 mode

### LINKED ISSUE #332: Blank nodes in list given invalid `.` at start in `text/n3` format.
Running

```
import { Parser } from 'n3';
const parser = new Parser({ format: 'text/n3' });
console.log(parser.parse('<s> <p> (_:a).'))
```

Creates an invalid blank node id of `_:.a`

### LINKED ISSUE #660: n3 mode: document-labelled blank node referenced inside a property list gets a graph-scoped label, splitting the node
## Bug

In `text/n3` mode, a document-labelled blank node (`_:c`) that is *referenced inside a `[ … ]` blank node property list* is given a graph-scoped label (`<currentGraphLabel>.c`) instead of the document-wide label — splitting what should be one blank node into two distinct nodes. At the top level the current graph label is empty, so the label is literally `.c`.

## Repro (n3 2.1.0, Node 22)

```js
const { Parser } = require('n3');
const doc = `@prefix : <http://example.org/>.
_:c :name "shared".
[ :ref _:c ] :note "property list".`;
const quads = new Parser({ format: 'text/n3', baseIRI: 'http://example.org/doc' }).parse(doc);
// quads:
//   _:b0_c  :name "shared"
//   _:n3-0  :ref  _:.c        <-- should be _:b0_c
//   _:n3-0  :note "property list"
```

Observed blank-node labels: `b0_c`, `n3-0`, **`.c`** — the two `_:c` mentions no longer co-refer.
Expected: both `_:c` mentions map to the same node (as in default/Turtle mode, where the same input yields exactly `b1_c` + `n3-1`).

Formula-internal scoping appears correct (inside `{ … }` the formula label is the right scope) — the defect is specifically the property-list path picking up the graph label for *document*-labelled nodes.

## Context

Found while migrating rdflib.js onto the N3.js parser (linkeddata/rdflib.js#831). Workaround used there: pass an explicit per-parse `blankNodePrefix` and rename the `.`-prefixed labels back to the document scope after parsing — works, but every n3-mode consumer would need to replicate it.

*(Agent-assisted report for @jeswr; repro verified against n3@2.1.0 before filing.)*

jeswr:
In N3 mode, `_saveContext` rescopes blank node labels and quantifiers on every context push — including lists and blank node property lists, not just formulae. Entering a top-level list resets the label prefix to `.`, so `<s> <p> (_:a).` produces the invalid label `_:.a` and no longer co-references `_:a` elsewhere in the document.

As this changes observable output, the fix is opt-in: the new `formulaScopedBlankNodes` parser option (name open to bikeshedding) rescopes labels and quantifiers only when entering a formula, matching N3's formula-scoped blank node semantics. The default preserves the released behaviour exactly, now pinned by explicit tests; #630 tracks flipping the default in the next major version.

With the option enabled, blank nodes inside `<<`/`<<(` contexts in N3 mode also share the enclosing scope, which matches their non-formula semantics — flagging this in case it needs separate discussion.

Closes #332 (via the option; the default flips in #630).

cc @jeswr

Closes #660

### DIFF
--- a/src/N3Parser.js
+++ b/src/N3Parser.js
@@ -87,11 +87,14 @@ export default class N3Parser {
     if (n3Mode) {
       // Every new scope resets the predicate direction
       this._inversePredicate = false;
-      // In N3, blank nodes are scoped to a formula
-      // (using a dot as separator, as a blank node label cannot start with it)
-      this._prefixes._ = (this._graph ? `${this._graph.value}.` : '.');
-      // Quantifiers are scoped to a formula
-      this._quantified = Object.create(this._quantified);
+      // In N3, blank nodes and quantifiers are scoped to a formula;
+      // lists and blank node property lists share the enclosing scope
+      if (type === 'formula') {
+        // Label the scope with the formula's blank node
+        // (using a dot as separator, as a blank node label cannot start with it)
+        this._prefixes._ = `${this._graph.value}.`;
+        this._quantified = Object.create(this._quantified);
+      }
     }
   }
 
--- a/src/N3Parser.js
+++ b/src/N3Parser.js
@@ -37,6 +37,8 @@ export default class N3Parser {
     this._lexer = options.lexer || new N3Lexer({ lineMode: isLineMode, n3: isN3, isImpliedBy: this._isImpliedBy });
     // Disable explicit quantifiers by default
     this._explicitQuantifiers = !!options.explicitQuantifiers;
+    // Disable formula-only blank node scoping by default
+    this._formulaScopedBlankNodes = !!options.formulaScopedBlankNodes;
     // Disable parsing of unsupported versions by default
     this._parseUnsupportedVersions = !!options.parseUnsupportedVersions;
     this._version = options.version;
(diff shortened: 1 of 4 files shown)


src/N3Parser.js
```diff
@@ -87,11 +87,14 @@ export default class N3Parser {
...
-      // Quantifiers are scoped to a formula
-      this._quantified = Object.create(this._quantified);
+      // In N3, blank nodes and quantifiers are scoped to a formula;
+      // lists and blank node property lists share the enclosing scope
+      if (type === 'formula') {
+        // Label the scope with the formula's blank node
+        // (using a dot as separator, as a blank node label cannot start with it)
+        this._prefixes._ = `${this._graph.value}.`;
+        this._quantified = Object.create(this._quantified);
+      }
```
jeswr, line 100: This constitutes a breaking change and should be opt-in until the next major version. If this lands with the optional behavior a ticket needs to be created to change the default in the next major version.

jeswr: @josd is this this the agreed upon / consensus behaviour in the N3 CG?

jeswr: Reworked in cbaeff1: the formula-only scoping is now gated behind a new opt-in parser option, `formulaScopedBlankNodes` (naming suggestions welcome). The default preserves the released behaviour exactly and is pinned by explicit tests; #630 tracks flipping the default in the next major version.

PR #616, opened 2026-07-02
fix: add opt-in @base directive writing via writeBase

### LINKED ISSUE #375: @base set by baseIRI is not shown in the prefixes part of the document
After setting baseIRI in the Writer constructor, the @base does not show up among the prefixes at the beginning of the generated turtle format, as the actions for it to display the @base are missing from the file.

Example: Having this in my code:

`new N3.Writer({ prefixes: { sh: 'http://www.w3.org/ns/shacl#', rdfs: "http://www.w3.org/2000/01/rdf-schema#"}, baseIRI: this.baseURL  })`

and the resulting file shows only this at the beginning, missing @base:

@prefix sh: <http://www.w3.org/ns/shacl#>.
@prefix rdfs: <http://www.w3.org/2000/01/rdf-schema#>.

jeswr:
When `baseIRI` is set, the writer relativises IRIs against it but never emits an `@base` directive, so the output only reads back correctly when consumed at exactly that base URL. With the new opt-in `writeBase: true` option, the writer writes `@base <baseIRI>.` ahead of the prefix declarations (Turtle/TriG only; N-Triples and N-Quads stay directive-free), so a round-trip through the parser restores the original IRIs regardless of where the document lands — the relativisation matrix verifies this by re-parsing without a base IRI.

Per review, the directive is opt-in: the default remains today's released output (no `@base` line, pinned by tests), and #631 tracks flipping the default in the next major version. The option name `writeBase` is open to bikeshedding.

Formatting/edge cases (simpler now that it's opt-in):
- `@base` joins the directive block, so no extra blank line is introduced: the blank line still only follows `@prefix` lines, and an `@base`-only block runs directly into the triples.
- When opted in, `@base` is written even for a `baseIRI` that cannot relativise some (or any) IRIs — absolute IRIs are unaffected by `@base`, so the output stays correct without special-casing.

Closes #375 — full resolution by default lands via #631.

cc @jeswr


src/N3Writer.js
```diff
@@ -58,10 +58,13 @@ export default class N3Writer {
       this._lineMode = false;
       this._graph = DEFAULTGRAPH;
       this._prefixIRIs = Object.create(null);
-      options.prefixes && this.addPrefixes(options.prefixes);
       if (options.baseIRI) {
         this._baseIri = new BaseIRI(options.baseIRI);
+        // Write the base as a directive, so relative IRIs
+        // resolve to the original IRIs regardless of document location
+        this._write(`@base <${options.baseIRI}>.\n`);
```
jeswr: This constitutes a breaking change and should be opt-in until the next major version. If this lands with the optional behavior a ticket needs to be created to change the default in the next major version.

jeswr: Done in f356621: `@base` emission is now opt-in via `writeBase: true` (name open to bikeshedding); the default keeps the released output, pinned by tests. #631 tracks flipping the default in the next major.

src/N3Writer.js
```diff
@@ -58,10 +58,15 @@ export default class N3Writer {
       this._lineMode = false;
       this._graph = DEFAULTGRAPH;
       this._prefixIRIs = Object.create(null);
-      options.prefixes && this.addPrefixes(options.prefixes);
       if (options.baseIRI) {
         this._baseIri = new BaseIRI(options.baseIRI);
+        // With the `writeBase` option, write the base as a directive,
+        // so relative IRIs resolve to the original IRIs
+        // regardless of document location
```
jeswr:
```suggestion
```

README.md
```diff
@@ -230,6 +230,16 @@ const writer1 = new N3.Writer({ format: 'N-Triples' });
 const writer2 = new N3.Writer({ format: 'application/trig' });
 ```
 
+A `baseIRI` argument makes the writer abbreviate IRIs relative to that base.
+<br>
+Pass `writeBase: true` to also write the base as an `@base` directive,
+so the abbreviated IRIs resolve to the original IRIs
+regardless of where the document is stored:
```
jeswr:
```suggestion
A `baseIRI` argument makes the writer abbreviate IRIs relative to that base. Pass `writeBase: true` to also write the base as an `@base` directive at the top of the document.
```

Issue #356, opened 2023-06-03
Incorrect parsing of empty Notation3 graph terms


phochste:
The current N3 parser doesn't handle empty Notation3 graph terms well:

This Notation3 document:

```
@prefix : <urn:example:> .
@prefix log: <http://www.w3.org/2000/10/swap/log#> .

() log:onNegativeSurface { } .
```

is parsed as:

```
[
      Quad {
        id: '',
        _subject: NamedNode { id: 'http://www.w3.org/1999/02/22-rdf-syntax-ns#nil' },
        _predicate: null,
        _object: null,
        _graph: BlankNode { id: '_:n3-72' }
      },
      Quad {
        id: '',
        _subject: NamedNode { id: 'http://www.w3.org/1999/02/22-rdf-syntax-ns#nil' },
        _predicate: NamedNode {
          id: 'http://www.w3.org/2000/10/swap/log#onNegativeSurface'
        },
        _object: BlankNode { id: '_:n3-72' },
        _graph: DefaultGraph { id: '' }
      }
]
```

See also: https://github.com/jeswr/pretty-turtle/issues/45

jeswr:
Also see https://github.com/rdfjs/types/issues/37 regarding the handling of empty graph terms.

jeswr:
Confirmed, and root cause found: nothing resets `this._subject` when entering a formula, so an empty `{ }` reaches `_readFormulaTail` with the enclosing statement's stale subject and emits the garbage `(rdf:nil, null, null, _:formula)` quad you're seeing. `{} => true .` hits the same path with an uninitialised subject (as noted in #328) and then crashes re-serialisation. The fix is up in #618: the stale-subject/garbage-quad bug is fixed unconditionally, and the `{}` ≡ `"true"^^xsd:boolean` mapping (the direction the w3c-cg/N3 group settled on, backed by the spec tests) is opt-in until the next major — #632 tracks flipping the default. Note that by default `{}` then still parses to a plain blank node with nothing marking it as an (empty) graph term — that representation question is rdfjs/types#37 and stays out of scope here.

PR #617, opened 2026-07-02
fix: link graph terms in lists instead of dropping them

### LINKED ISSUE #342: list parsing issue
a list containing a graph apparently is ignored:

```js
import { Parser } from 'n3'
const parser = new Parser({ format: 'text/n3' })
console.log(parser.parse('<s> <p> ({<a> <b> <c>}) .'))
```

### LINKED ISSUE #357: Incorrect parsing of a list containing a Notation3 graph term
The current 1.16.4 N3 parser doesn't handle lists contains Notation3 graph terms well:

This Notation3 document:

```
@prefix : <urn:example:> .

:s a ({ :a :b :c}) .
```

is parsed as:

```
Quad {
  id: '',
  _subject: NamedNode { id: 'urn:example:a' },
  _predicate: NamedNode { id: 'urn:example:b' },
  _object: NamedNode { id: 'urn:example:c' },
  _graph: BlankNode { id: '_:n3-0' }
}
Quad {
  id: '',
  _subject: NamedNode { id: 'urn:example:s' },
  _predicate: NamedNode { id: 'http://www.w3.org/1999/02/22-rdf-syntax-ns#type' },
  _object: NamedNode { id: 'http://www.w3.org/1999/02/22-rdf-syntax-ns#nil' },
  _graph: DefaultGraph { id: '' }
}
```

jeswr:
An N3 graph term inside a list was silently dropped: the `{` branch of `_readListItem` pushed the formula context and returned before any of the list bookkeeping ran, so the list head was never allocated and no `rdf:first`/`rdf:rest` linkage was emitted — `<s> <p> ({<a> <b> <c>}) .` parsed as `<s> <p> rdf:nil` while the formula's quads floated unattached.

The fix models the `{` case on the `[` case: allocate the list cell, link the formula's blank node into the list, then enter the formula. The linkage cannot fall through to the shared code after the `switch`, because it must be emitted in the enclosing graph before `this._graph` switches to the formula. `_subject` is also cleared on entry so an empty `{}` item cannot emit the enclosing statement's stale subject at `}`; the `{}` ≡ `true` mapping itself is left to #356/#374.

Closes #342. Also resolves #357, which is the same defect.

cc @jeswr


test/N3Parser-test.js
```diff
@@ -2872,11 +2872,60 @@ describe('Parser', () => {
...
+      'should parse a formula between other list items',
+      shouldParse(parser, '<s> <p> (<x> {<a> <b> <c>} <y>) .',
+      ['s', 'p', '_:b0'],
+      ['_:b0', 'http://www.w3.org/1999/02/22-rdf-syntax-ns#first', 'x'],
+      ['_:b0', 'http://www.w3.org/1999/02/22-rdf-syntax-ns#rest', '_:b1'],
+      ['_:b1', 'http://www.w3.org/1999/02/22-rdf-syntax-ns#first', '_:b2'],
+      ['_:b1', 'http://www.w3.org/1999/02/22-rdf-syntax-ns#rest', '_:b3'],
+      ['_:b3', 'http://www.w3.org/1999/02/22-rdf-syntax-ns#first', 'y'],
+      ['_:b3', 'http://www.w3.org/1999/02/22-rdf-syntax-ns#rest', 'http://www.w3.org/1999/02/22-rdf-syntax-ns#nil'],
+      ['a', 'b', 'c', '_:b2'],
```
jeswr, line 3111: We need test cases with layers of complex nesting between graphs and lists.

jeswr: Added a deep-nesting matrix in 9b72292: list>formula>list (object and subject position, plus a 3-level list/formula alternation), a formula containing a list of formulae, a formula inside a nested list `((<x> {..}) <y>)`, formulae in subject- and object-position lists of one triple, adjacent formulae in a nested list, a formula-list inside a blank node property list, and `@forSome` interacting with a formula in a list — all expected quads derived by hand from the N3 list/formula semantics (linkage in the enclosing graph, contents in the formula graph).

phochste:
Any idea when this will arrive in a next release? I'm working on N3 code in JS and constantely hitting this bug.

PR #618, opened 2026-07-02
fix: stop empty N3 graph terms leaking garbage quads; opt-in "true"^^xsd:boolean parsing

### LINKED ISSUE #356: Incorrect parsing of empty Notation3 graph terms

jeswr:
Ports #374 by @Dexagod onto main and finishes it, reworked per review into an unconditional bug fix plus an opt-in behavior. The original patch targeted the superseded `versions/2.0.0` branch and no longer applies, so authorship is carried via `Co-authored-by`.

**Unconditional (the #356 bug fix):** `_saveContext` clears `_subject` when entering a formula, so the enclosing statement's subject no longer leaks in — that leak produced the garbage `(rdf:nil, null, null, _:formula)` quad reported in #356 and null-predicate quads that broke re-serialisation. An empty `{}` now always yields a clean blank-node graph term with no spurious quads.

**Opt-in (`emptyFormulaAsTrue`, default `false`):** parses an empty `{}` as `"true"^^xsd:boolean` instead. w3c-cg/N3#185 discusses this direction but has not settled it; the N3 spec tests are the operative guide, and they expect the boolean reading. Per review, the default flips only in the next major — tracked in #632. (Flag name open to bikeshedding; it follows the `isImpliedBy` precedent of #500.) Emptiness is tracked per formula context instead of inferred from `_predicate === null`, which only held for the first statement of a document.

**Spec tests:** the suite's two empty-graph cases ([`empty_graph_eval`](https://w3c-cg.github.io/N3/tests/N3Tests/manifest-parser.ttl#empty_graph_eval), [`empty_graph_implies_eval`](https://w3c-cg.github.io/N3/tests/N3Tests/manifest-parser.ttl#empty_graph_implies_eval), both `rdft:Approved`, added in w3c-cg/N3#202) are ported as unit tests, asserted under the option since they expect the boolean reading; the default blank-node behavior is pinned by separate tests. They are `TestN3Eval` cases whose results the suite states as documents, so the unit tests assert the exact quads N3.js emits (formula terms as blank nodes). Wiring the full N3 suite into CI remains #331.

Deliberately unchanged: formulae containing only quantifiers (`{ @forSome <x>. }`) stay blank-node terms, formulae in predicate position remain unsupported (empty or not), and empty graph terms inside lists still collapse — that is the separate list bug #342, addressed in #617.

Closes #356

cc @jeswr


jeswr:
I would not say a decision has been made in (https://github.com/w3c-cg/N3/issues/185). The spec tests _are_ probably the best guide that we can have as to what the default behaviour - we should also adopt those spec tests in this PR if approrpiate/possible.

The behaviour should be opt-in until the next major version to avoid breaking changes in a minor version. Please open a ticket tagged next major version so that we do flip this behavior in the next mver if all maintainers agree to.

jeswr:
Reworked in ba6b9a2:

- **Split the change.** The `_subject`-leak fix stays unconditional (it is the actual #356 bug: garbage `(rdf:nil, null, null, _:formula)` quads and null-predicate quads that broke re-serialisation) — an empty `{}` now always yields a clean blank-node graph term by default. The `{}` → `"true"^^xsd:boolean` mapping is now opt-in behind a new `emptyFormulaAsTrue` parser option (default `false`), following the `isImpliedBy` precedent from #500. Name open to bikeshedding.
- **Softened the w3c-cg/N3#185 wording** in the PR body, code comments, and tests: it is a discussion direction, not a decision; the spec tests are the operative guide.
- **Adopted the spec tests.** The suite has two approved empty-graph cases, [`empty_graph_eval`](https://w3c-cg.github.io/N3/tests/N3Tests/manifest-parser.ttl#empty_graph_eval) and [`empty_graph_implies_eval`](https://w3c-cg.github.io/N3/tests/N3Tests/manifest-parser.ttl#empty_graph_implies_eval) (from w3c-cg/N3#202); both are ported as unit tests, cited by URL. They expect the boolean reading, so they are asserted under the option; the default blank-node behaviour is pinned by separate tests. Full-suite CI wiring stays with #331.
- **Next-major flip** is tracked in #632 (tagged `semver.major`), for flipping the default if all maintainers agree.

PR #623, opened 2026-07-02
docs: note prefix/comment event ordering on StreamParser

### LINKED ISSUE #440: Add warning that StreamParser prexied and comment events are not synchronized to data
https://github.com/rdfjs/N3.js/pull/423 introduced the `comment` event on `StreamParser`. In https://github.com/rdfjs/N3.js/pull/423, I realized that `comment` emitted by `StreamParser` is not guaranteed to be synchronized. That is, there is no guarantee that a timeline of `quad comment quad` means that the comment in question was actually positioned in between the two quads in the source file.

Similarly, the `prefix` event is also not reliable.

This is inherently unfixable with `ReadStream<Quad>` (only fixable with `ReadStream<Quad|Prefix|Comment>`), so we should document this. Consumers MUST NOT rely on synchronization, because there is none.

This is less harmful for usage of `prefix` and `comment` for non-location-specific information purposes (as would nearly always be the case for prefixes). If any `prefix` or `comment` handling pertains to a specific position, `StreamParser` usage must be replaced by `Parser`, which does provide order guarantees.

jeswr:
On `N3.StreamParser`, `prefix` and `comment` events are emitted as soon as they are parsed, while quads can remain buffered in the Transform's readable queue, so their order relative to `data` events need not match the document. Verified empirically: with the README's SlowConsumer pattern, a later prefix and trailing comment were observed before earlier quads were delivered. This adds a warning paragraph to the README pointing users who need document-order positioning to `N3.Parser` with `onQuad`/`onPrefix`/`onComment`, whose callbacks were verified to fire in document order.

Closes #440

cc @jeswr

### DIFF
--- a/README.md
+++ b/README.md
@@ -204,6 +204,14 @@ function SlowConsumer() {
 A dedicated `prefix` event signals every prefix with `prefix` and `term` arguments.
 A dedicated `comment` event can be enabled by setting `comments: true` in the N3.StreamParser constructor.
 
+Note that `prefix` and `comment` events are emitted as soon as they are parsed,
+whereas quads can remain buffered until the consumer is ready to read them.
+The order of these events relative to `data` events is therefore
+not guaranteed to match the position of prefixes and comments in the document.
+If their position matters,
+use `N3.Parser` with the `onQuad`, `onPrefix` and `onComment` callbacks instead,
+which are invoked in document order.
+
 ## Writing
 
 ### From quads to a string
--- a/src/N3StreamParser.js
+++ b/src/N3StreamParser.js
@@ -1,6 +1,4 @@
 // **N3StreamParser** parses a text stream into a quad stream.
-// Because quads can remain buffered until the consumer reads them,
-// `prefix` and `comment` events are not synchronized with `data` events.
 import { Transform } from 'readable-stream';
 import N3Parser from './N3Parser';
 
--- a/src/N3StreamParser.js
+++ b/src/N3StreamParser.js
@@ -1,4 +1,6 @@
 // **N3StreamParser** parses a text stream into a quad stream.
+// Because quads can remain buffered until the consumer reads them,
+// `prefix` and `comment` events are not synchronized with `data` events.
 import { Transform } from 'readable-stream';
 import N3Parser from './N3Parser';
 


src/N3StreamParser.js
```diff
@@ -1,4 +1,6 @@
 // **N3StreamParser** parses a text stream into a quad stream.
+// Because quads can remain buffered until the consumer reads them,
+// `prefix` and `comment` events are not synchronized with `data` events.
```
jeswr: This comment does not seem appropriate here.

jeswr:
```suggestion
```

jeswr: Agreed — the note is removed from the class header in 8ab6071 (via the applied suggestion); the README warning remains. PR description updated to match.

README.md
```diff
@@ -204,6 +204,14 @@ function SlowConsumer() {
 A dedicated `prefix` event signals every prefix with `prefix` and `term` arguments.
 A dedicated `comment` event can be enabled by setting `comments: true` in the N3.StreamParser constructor.
 
+Note that `prefix` and `comment` events are emitted as soon as they are parsed,
+whereas quads can remain buffered until the consumer is ready to read them.
+The order of these events relative to `data` events is therefore
+not guaranteed to match the position of prefixes and comments in the document.
+If their position matters,
+use `N3.Parser` with the `onQuad`, `onPrefix` and `onComment` callbacks instead,
+which are invoked in document order.
```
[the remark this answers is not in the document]
jeswr: Resolved: the class-doc note was removed per maintainer review (8ab6071), and the PR description has been updated to drop that claim — the change is now README-only.

jeswr:
This is a controversial docs update. Merging without additional review.

TallTed:
> This is a controversial docs update. Merging without additional review.

I sincerely hope that the above was meant to say --

> This is _**an uncontroversial**_ docs update.

jeswr:
> > This is a controversial docs update. Merging without additional review.
> 
> I sincerely hope that the above was meant to say --
> 
> > This is _**an uncontroversial**_ docs update.

Yes indeed it was!

Issue #453, opened 2024-09-21
Enable jest eslint rules


jeswr:
Per https://github.com/rdfjs/N3.js/pull/452/files#r1769650058

jeswr:
Partial progress here: eslint-plugin-jest's recommended config now applies to test/** (since #452, carried through the flat-config migration in #600), and jest/no-conditional-expect is enabled. Still disabled: jest/no-standalone-expect, jest/expect-expect, jest/no-done-callback, jest/no-identical-title. All four are mechanical — expect-expect/no-standalone-expect mostly need assertFunctionNames for our shouldParse-style helpers, no-identical-title needs a title-dedupe pass, and no-done-callback means converting roughly 37 done-style tests to promises. Best landed as one small PR per rule. jest/expect-expect and jest/no-identical-title are up in #624 and #629; no-done-callback (~37 tests) remains.

PR #624, opened 2026-07-02
chore(tests): enable jest/expect-expect

jeswr:
Enables `jest/expect-expect` in the test ESLint config, listing the suites' shared assertion helpers (`shouldParse`, `shouldTokenize`, `shouldSerialize`, …) in `assertFunctionNames`. The rule surfaced three tests that asserted nothing; they now assert the behaviour their titles describe (no-throw for the lexer/parser cases, and that a failing output-stream `end` is swallowed so the writer's `done` callback is called without an error).

`jest/no-standalone-expect` stays disabled for now: N3Store-test.js asserts in `describe` bodies and `beforeAll` hooks while building up shared mutable stores, and enabling it means restructuring those suites — better as its own change.

Refs #453

cc @jeswr

### DIFF
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -316,9 +316,15 @@ export default [
         ],
       }],
 
-      // Still disabled: N3Store-test.js asserts in `describe` bodies and
-      // `beforeAll` hooks while building up its shared mutable stores;
-      // enabling this requires restructuring those suites.
+      // The `0` entries below override rules that `jest.configs['flat/recommended']`
+      // (spread into the preceding `test/**` config object) enables as `error`;
+      // in flat config, the later entry for the same files wins.
+      //
+      // `jest/no-standalone-expect` stays off — independently of
+      // `jest/expect-expect` above — because N3Store-test.js has ~40
+      // standalone `expect`s in `describe` bodies and `beforeAll` hooks
+      // while building up its shared mutable stores; enabling it requires
+      // restructuring those suites.
       'jest/no-standalone-expect': 0,
       'jest/no-done-callback': 0,
       'jest/no-identical-title': 0,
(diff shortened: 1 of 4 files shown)


eslint.config.mjs
```diff
@@ -295,8 +295,31 @@ export default [
...
+          'shouldParse',
+          'shouldParseWithCommentsEnabled',
+          'shouldSerialize',
+          'shouldTokenize',
+        ],
+      }],
+
+      // Still disabled: N3Store-test.js asserts in `describe` bodies and
+      // `beforeAll` hooks while building up its shared mutable stores;
+      // enabling this requires restructuring those suites.
```
jeswr: I don't understand how this is being disabled.

jeswr: The rule arrives enabled: `jest.configs['flat/recommended']` (spread into the preceding `test/**` config object) sets `jest/no-standalone-expect` to `error`, and in flat config a later entry for the same files wins — so the `0` here switches it back off. I verified it's live config: deleting the line yields 40 `no-standalone-expect` errors in `test/N3Store-test.js` (standalone `expect`s in `describe` bodies / `beforeAll` hooks), which is also why it stays off, independently of `jest/expect-expect`. Made the config say this explicitly in e6fa4f6.

eslint.config.mjs
```diff
@@ -295,8 +295,37 @@ export default [
...
+
+      // The `0` entries below override rules that `jest.configs['flat/recommended']`
+      // (spread into the preceding `test/**` config object) enables as `error`;
+      // in flat config, the later entry for the same files wins.
+      //
+      // `jest/no-standalone-expect` stays off — independently of
+      // `jest/expect-expect` above — because N3Store-test.js has ~40
+      // standalone `expect`s in `describe` bodies and `beforeAll` hooks
+      // while building up its shared mutable stores; enabling it requires
+      // restructuring those suites.
```
jeswr:
```suggestion
```

Issue #246, opened 2021-04-26
DefaultGraph as subject instead of NamedNode for relative baseIRI <>


spetrac-nicos-ag:
I am currently implementing a ldp-server. I am using n3 to load a turtle, like in the examples of solid rest api (https://github.com/solid/solid-spec/blob/master/api-rest.md#creating-containers-directories).
```turtle
<> <http://purl.org/dc/terms/title> "Basic container" .
```

In this case, the `<>` indicates the relative baseIRI (location) for the POST request and must be loaded as NamedNode. The N3Parser on the other hand parses these as DefaultGraph, because it just organizes keys and the defaultGraph has the same key "" as the NamesNode <>. The problem appears in the rendering of the dataset back to turtle with the N3Writer, because the output of the DefaultGraph is an empty string and no longer the `<>`.
```turtle
 <http://purl.org/dc/terms/title> "Basic container" .
```

This results in an invalid turtle, although it has been parsed from valid turtle and just serialized as shown:
```js
const turtle   = '<> <http://purl.org/dc/terms/title> "Basic container" .';
const store    = new N3.Store();
const prefixes = await new Promise((resolve, reject) => {
    const parser = new N3.Parser();
    parser.parse(turtle, (error, quad, prefixes) => {
        if (error) reject(error);
        else if (quad) store.addQuad(quad);
        else resolve(prefixes);
    });
});
const writer   = new N3.Writer({prefixes});
store.getQuads().forEach(quad => writer.addQuad(quad));
const result = await new Promise((resolve, reject) => writer.end(
    (err, result) => err ? reject(err) : resolve(result)
));
```

I think the DefaultGraph should never appear in any other position than the graph position and should be returned as NamedNode otherwise. This would solve the issue with the turtle rendering immediatly. as shown in this quick and dirty fix:
```js
//store.getQuads().forEach(quad => writer.addQuad(quad));
store.getQuads().forEach((quad) => {
    if (quad.subject.termType === 'DefaultGraph') {
        writer.addQuad(N3.DataFactory.quad(
            N3.DataFactory.namedNode(''),
            quad.predicate,
            quad.object,
            quad.graph
        ));
    } else {
        writer.addQuad(quad);
    }
});
```

RubenVerborgh:
> I think the DefaultGraph should never appear in any other position than the graph position and should be returned as NamedNode otherwise.

That is correct, thanks for reporting.

If you have the bandwidth for a PR, happy to review.

[...]

PR #625, opened 2026-07-02
fix: write empty named nodes as <>

jeswr:
A quad in graph `NamedNode('')` was silently merged into the default graph (the id-based fast path of `equals` treats the two as equal), and `quadToString` dropped such a graph via a `graph.value` truthiness check. Both now serialize the graph as `<>`, matching the existing behaviour for empty named nodes in the other three positions, so output stays valid Turtle/TriG and round-trips.

Refs #246 — this covers only the writer layer; the store still conflates `NamedNode('')` with the default graph (both use `''` as internal id), which is left for a separate change.

cc @jeswr

### DIFF
--- a/src/N3Writer.js
+++ b/src/N3Writer.js
@@ -85,7 +85,9 @@ export default class N3Writer {
   _writeQuad(subject, predicate, object, graph, done) {
     try {
       // Write the graph's label if it has changed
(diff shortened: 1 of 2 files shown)


jeswr (commented):
This is a sensible patch. Approved.

PR #627, opened 2026-07-02
feat: let extractLists tolerate lists with extra arcs

### LINKED ISSUE #546: N3Store.exractLists() too restrictive
Currently, N3 assumes that RDF lists consist of quads having a predicate of `rdf:first` or `rdf:rest` **only**. Is there any reason for this?

I would expect this example to be valid, but N3 declares it as malformed with the error `http://example.org/1 can't be subject and object`:

```
@prefix rdf: <http://www.w3.org/1999/02/22-rdf-syntax-ns#> .
@prefix foaf: <http://xmlns.com/foaf/0.1/> .
@prefix : <http://example.org/> .

:alice foaf:name  "Alice" .
:bob   foaf:name  "Bob" .
:cal   foaf:knows :1 .

:1 rdf:first :alice ; rdf:rest :2 .
:2 rdf:first :bob   ; rdf:rest rdf:nil .

# as soon as we add other triples to the list, it is declared as malformed
:1 a rdf:List .
```

See e.g. [this ontology](https://github.com/tibonto/aeon/blob/main/aeon-full.owl#L2142) for a real world example that we would like to use together with N3.

jeswr:
`extractLists()` currently declares a list malformed as soon as one of its nodes carries any arc besides `rdf:first`/`rdf:rest` (e.g. `:1 a rdf:List`), which is common in real-world ontologies.

This adds an opt-in `allowExtraArcs` option: such lists are returned in the mapping, but their quads are never pushed to the removal set, so with `remove: true` they are left fully intact (like malformed lists today) and nothing dangles. The default stays strict on purpose: the result typically feeds the writer's `lists` option, and silently dropping extra arcs there would make round-tripping lossy — with the option enabled, callers accept that remaining `rdf:first`/`rdf:rest` quads coexist with the returned mapping.

Also improves the error the reporter hit: `:1 can't be subject and object` now reads `:1 has non-list arc http://...ns#type`, naming the offending predicate.

Note: a list head that has no incoming arcs and carries both a subject-position use and an extra arc remains an error even with the option, since it is ambiguous which arc references the list.

Naming is up for debate — `allowExtraArcs` vs e.g. `tolerateExtraArcs`/`lenient`; happy to rename.

Closes #546

cc @jeswr

### DIFF
--- a/src/N3Store.js
+++ b/src/N3Store.js
@@ -758,20 +758,26 @@ export default class N3Store {
 
   // ### `extractLists` finds and removes all list triples
   // and returns the items per list.
-  extractLists({ remove = false, ignoreErrors = false } = {}) {
+  // With `allowExtraArcs`, lists whose nodes carry arcs other than
+  // rdf:first/rdf:rest are returned as well,
+  // but left intact when `remove` is set.
+  extractLists({ remove = false, ignoreErrors = false, allowExtraArcs = false } = {}) {
     const lists = {}; // has scalar keys so could be a simple Object
     const onError = ignoreErrors ? (() => true) :
                   ((node, message) => { throw new Error(`${node.value} ${message}`); });
 
     // Traverse each list from its tail
     const tails = this.getQuads(null, namespaces.rdf.rest, namespaces.rdf.nil, null);
-    const toRemove = remove ? [...tails] : [];
+    const toRemove = [];
     tails.forEach(tailQuad => {
       const items = [];             // the members found as objects of rdf:first quads
       let malformed = false;      // signals whether the current list is malformed
+      let extraArcs = false;      // signals whether the current list has extra arcs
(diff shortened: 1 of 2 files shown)


jeswr (commented):
LGTM

src/N3Store.js
```diff
@@ -843,9 +853,14 @@ export default class N3Store {
...
-      // Store the list under the value of its head
-      else if (head)
-        lists[head[headPos].value] = items;
+      else {
+        // Store the list under the value of its head
+        if (head)
+          lists[head[headPos].value] = items;
+        // Leave lists with extra arcs fully intact
+        if (extraArcs)
+          toRemove.length = removeStart;
```
jeswr: Setting the length of a list is very odd behavior. I don't understand what is happening here.

PR #628, opened 2026-07-02
fix: accept lists in the predicate position in N3 mode

### LINKED ISSUE #619: N3 mode does not accept lists in the predicate position
In Notation3, a predicate can be any expression, including a list — but the parser only accepts lists in the subject and object positions:

```n3
(<b> <c>) <a> <d>.    # parses
<a> <b> (<c> <d>).    # parses
<a> (<b> <c>) <d>.    # Expected entity but got ( on line 1
```

(verified on n3@2.1.0 with `format: 'text/n3'`)

This was implemented on the superseded `versions/2.0.0` branch in #345 (a `(` case in `_readPredicate`, gated on N3 mode) but never made it to `main` — see the #341 close-out. It should be re-implemented against current `main` as a small focused PR, and validated against the W3C N3 test suite once #331 lands.

Related N3-mode list fixes in flight: #342/#357 (graph terms in lists, #617) and #328 (literal subjects/predicates, #613).

jeswr:
In N3 mode, `_readPredicate` rejected `(`, so `<a> (<b> <c>) <d>.` failed with `Expected entity but got (` while lists in the subject and object positions parsed fine. This extracts the shared list-start logic into a `_readList` helper, allows `(` as a predicate in N3 mode only (Turtle/TriG still reject it with the same error as before), and teaches the list-close/first-element paths to install the list head as the predicate and continue into the object.

Ports #345 from the superseded `versions/2.0.0` branch onto `main`; the surrounding code has changed since (reified-triple guards, `_readListItem`), so this re-applies the intent rather than the patch.

Note: small expected rebase overlap with in-flight #613 (`_readPredicate`) and #617 (`_readListItem`) — whichever lands last rebases.

Closes #619

cc @jeswr

### DIFF
--- a/test/N3Parser-test.js
+++ b/test/N3Parser-test.js
@@ -1,4 +1,4 @@
-import { Parser, NamedNode, BlankNode, Quad, termFromId, DataFactory as DF } from '../src';
+import { Parser, Writer, NamedNode, BlankNode, Quad, termFromId, DataFactory as DF } from '../src';
 import rdfDataModel from '@rdfjs/data-model';
 import { isomorphic } from 'rdf-isomorphic';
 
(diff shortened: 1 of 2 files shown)


test/N3Parser-test.js
jeswr, line 1: We should add some more fuzzer style tests which have nested lists with multiple layers across each position

jeswr: Added in a698d53: (1) hand-derived deep cases — nested lists in subject+predicate+object simultaneously (`((<a>) (<b> (<c>))) ((<p>)) ((<x> (<y>)))`), a triple-nested predicate list, a nested predicate list inside a formula, empty lists at each depth/position, and lists in all three positions of a statement inside a formula; (2) a deterministic table-driven suite covering every subject/predicate/object depth combination 0-3 (63 documents), asserting quad-count formulas, exactly one rdf:first + rdf:rest arc per cell, nil-terminated unshared chains reachable from the main statement, and a Writer round-trip checked with rdf-isomorphic. One note: list-as-predicate inside a formula *inside a list* currently collapses the outer list to rdf:nil on this branch — that is the formula-in-list bug #617 fixes, so the combined construct only works once both PRs land; happy to add that test to whichever merges second.

PR #622, opened 2026-07-02
docs: document validation scope and external term validation

jeswr:
Recurring issues (#308, and the writer side in #329/#383) stem from the same expectation gap: N3.js validates per-format *grammar*, not the terms themselves. This adds a short **Validation** subsection under *Compatibility* stating what the parser guarantees, what it deliberately does not check (RFC 3987 well-formedness beyond the grammar, unresolved relative IRIs without `baseIRI`, literal value vs datatype, BCP 47 language tags, and that the writer trusts its input terms), and a compact validating-Transform snippet composing `validate-iri`, `rdf-validate-datatype`, and `bcp-47` for users who need term validation.

All claims and the snippet were verified against n3@2.1.0 and current `main` (e.g. `<http://example.org/%ZZ>` and `"abc"^^xsd:integer` parse; the snippet rejects both).

Refs #308

cc @jeswr

PR #634, opened 2026-07-02
feat: opt-in validation modes for the parser

### LINKED ISSUE #666: Validate blank-node labels and literal base directions in opt-in parser validation (#634)
The opt-in `validate` option in #634 checks the well-formedness of the terms the parser produces *after* parsing — IRIs (RFC 3987), language tags (BCP 47), and the lexical space of the common XSD datatypes. Two term dimensions are still unchecked and would round out `validate: { terms: true }`:

- **Blank-node labels** — validate the label against the blank-node-label production (`PN_CHARS_BASE` / `PN_CHARS` with the usual first/last-character rules), so a label carrying characters that have *no escape mechanism* in the Turtle family — whitespace, `<`, `>`, `.`, control characters — is rejected rather than carried through to serialization. The default lexer already constrains the labels it mints, so this is a defence-in-depth / round-trip-safety check for terms that reach the parser's `factory` (or a downstream writer) from non-grammar sources.
- **Literal base direction** — where a directional language-tagged literal is produced, validate that the direction is exactly `ltr` or `rtl`. This should compose with the existing `version` profile: directional language tags are an RDF 1.2 feature, so under `version: '1.1'` they are already rejected by `validate: { version: true }`; this check covers well-formedness of the direction value itself where it *is* permitted.

### Suggested implementation

Both checks belong alongside the existing ones in `src/Validation.js` (e.g. `isValidBlankNodeLabel`, `isValidBaseDirection`), wired into the same per-instance method overrides so there is still **zero cost when `validate` is off**. Tests go in `test/Validation-test.js`, and the `validate` section of the README should list the two added dimensions.

### Background

These are the two remaining term dimensions from a term-well-formedness review (tracked there as **M4 — blank-node labels** and **M5 — language tag / base direction**; the language-tag half is already covered by #634). They are input-well-formedness / GIGO concerns rather than default-hot-path guards, so the opt-in validation layer added in #634 is their natural home rather than a fail-closed throw in the writer's default path.

jeswr:
Adds an opt-in `validate` option to `N3.Parser`, as requested in https://github.com/rdfjs/N3.js/pull/622#discussion_r3514530718. Refs #622.

Design: the option covers the validation dimensions that `format` does not already gate.

- `validate: { terms: true }` — well-formedness of the *terms* the grammar produces: IRIs must be absolute and match RFC 3987's `IRI` rule (rejecting unresolved relative IRIs, `<http://example.org/%ZZ>`, …), language tags must be well-formed per BCP 47, literal values of `xsd:boolean|integer|decimal|double|float|date|time|dateTime` must be in their lexical space, blank-node labels must match the `BLANK_NODE_LABEL` production (astral-aware), and base directions must be `ltr`/`rtl`. Lexical checks only — no registry lookups, no value-space rules — implemented internally (`src/Validation.js`) to keep N3.js dependency-free. (Blank-node labels and base direction were folded in from #666/M4+M5 in 004d9b3.)
- `validate: { version: true }` — profile enforcement composed with the existing `version` option: `version: '1.1'` rejects triple terms, reified triples, annotation syntax, directional language tags, and version declarations; `version: '1.2-basic'` rejects triple terms and their sugar (per RDF 1.2 Concepts, Basic excludes only triple terms).
- `validate: true` enables both.

Deliberately not covered: RDF/JS interface conformance (the parser constructs terms through its `factory`, so that check would only re-test the factory — it belongs at the factory boundary); N3 validity (already what `format: 'text/n3'` does); shape-level validation (separate layer, noted in the #622 docs).

**Zero cost when off**: with `validate` absent the parser runs identical code — validation is installed as per-instance method overrides in the constructor, so no check was added to the default read path. Measured within noise of `main` (a control build with the checks stripped was no faster). Validation **on** is ~40 % slower on a 200 k-triple parse, dominated by the RFC 3987 regex. The only default-path additions are two guards on rare paths that also fix a latent crash: `_completeSubjectLiteral` previously dereferenced the result of `_completeLiteral` without the undefined check its object-position counterpart has.

Reviewer notes on the 004d9b3 additions: (1) in N3 mode only the written `token.value` of a blank-node label is checked (N3 deliberately scopes dot-prefixed internal labels), so a malformed `blankNodePrefix` is not caught in N3 format specifically — pinned by a regression test; (2) the direction check is defense-in-depth (the built-in lexer only emits `ltr`/`rtl`; it bites only for a custom `options.lexer`); (3) direction validity is enforced in all versions under `validate.terms`; version-gating of directional tags stays with `validate.version`.

Full suite green — 6666 tests, 100 % coverage.

cc @jeswr — design-discussion draft; the option shape (`validate: true` / `{ terms, version }`) and the datatype list are the parts most worth debating.

Closes #666


jeswr:
Folded the two remaining term dimensions from #666 (M4 blank-node labels, M5 base direction) into `validate: { terms: true }` in 004d9b3; details and the three reviewer notes now live in the PR description. Suite green — 6666 tests, 100 % coverage. (Prepared with Claude; I'll review before it progresses.)

Issue #633, opened 2026-07-02
Index RDF 1.2 triple terms and match patterns inside reified quads


jeswr:
Two related improvements to how the store handles RDF 1.2 triple terms:

1. **Index triple terms instead of dictionary-encoding them.** Today a triple term is stored in the entity index as a composite string id built from its component ids (`.s.p.o`), i.e. it lives in the dictionary like an opaque atom. That makes every structural operation (decomposition, component lookup, matching by component) a string-parsing exercise, and interacts awkwardly with cross-index operations. A dedicated indexing structure for triple terms — mapping component ids to the term's id the same way the quad indexes map term ids to quads — would make their structure first-class.

2. **Extend `match()` to patterns inside reified quads.** With the above, `match`/`getQuads` can support patterns where a `Quad` term contains wildcards — e.g. find all `?r rdf:reifies <<( :s ?p ?o )>>` statements by matching on the triple term's subject only, without a full scan. Today a `Quad` passed as a pattern term only ever matches by exact composite id.

Design questions to settle in the PR: how wildcards inside a nested quad pattern are expressed (`Variable` components being the natural RDF/JS candidate), recursive nesting, and the compatibility story for `termToId`/`termFromId`, the exported `EntityIndex` (#394), and the in-flight cross-index set operations (#601). Performance requirement: no regression on workloads without triple terms.

PR #635, opened 2026-07-02
perf: index triple terms by component ids instead of composite strings

jeswr:
First half of #633, per the two-PR split from the design discussion: swaps the entity index's representation of triple terms from composite dictionary strings (`.s.p.o[.g]` entries in `_ids`/`_entities`) to a numeric-id index, with no change in matching semantics. Wildcard matching inside `Quad` patterns (`<<( :s ?p ?o )>>`) builds on this index in the follow-up PR (#637).

- `N3EntityIndex` gains `_quadIds`, a three-level forward map `sId → pId → oId → termId`. The object level stores the term id directly for default-graph triple terms (the only kind the parser produces) and spills to a graph sub-map (`graphId → termId`, `1` = default graph) only when a graph-component term is interned for the same s/p/o, so the common case stays three levels deep. (A uniform fourth graph level measured +35–47 % heap on the triple-term workload; the spill layout is near parity.)
- The reverse entry for a triple term is its component-id array: `_entities[termId] = [sId, pId, oId(, gId)]`; `_ids` no longer contains composite entries. Term ids come from the same counter as before, preserving the ascending components-before-composite invariant that `remapEntityIds` relies on.
- `_remapCompositeId` becomes an array walk plus a forward-map probe (no string allocation); `_termFromId` materializes by component-id recursion instead of `split('.')`.
- Quad patterns still resolve to a single term id (exact match only); plain-quad code paths are untouched.

**Stacked on #601**: the diff includes #601's commits until it merges; I'll rebase afterwards. #601's `_remapCompositeId` unit tests are adapted to the array form (they probed the dot-string format directly); all other existing tests are unchanged.

Performance (M1, Node 25, babel-built `lib/`, fresh process per run, interleaved base/PR pairs vs #601 head bdf4213; triple-term workload `ex:r{i} rdf:reifies <<( ex:s{i%K} ex:p{i%16} ex:o{i} )>>` at 1 M quads): triple-term ingest 6 974 → 5 542 ms (faster in all three pairs); exact quad-term match 3.07 → 2.47 µs/op; `_ids` entries 3 010 018 → 2 010 018. Triple-term heap 1 029.46 → 1 037.81 B/quad (+0.8 %; +11 % at 100 k). Plain-quad ingest and heap are unchanged (per-quad heap identical; timing within noise). RDF 1.2 syntax suites (`spec-1-2-*`) pass; full test suite green at 100 % coverage.

Refs #633.

cc @jeswr


jeswr:
Can you please create an architecture diagram visualising what is happening here.

TallTed:
> [@jeswr] Can you please create an architecture diagram visualising what is happening here.

I think you might need to address the bot (or human) you're talking to.

Issue #435, opened 2024-09-05
Bug: `import`, `some`, `forEach` and `every` do not fully align with `Dataset` interface


jeswr:
Discovered when trying to update `@rdfjs/types` (attempt [here](https://github.com/jeswr/DefinitelyTyped/pull/new/wip/implement-dataset)). The error log when trying to make the `Store` implement the `Dataset` interface is:

```bash
  288:5  error  TypeScript@4.8, 4.9, 5.0, 5.1, 5.2, 5.3, 5.4, 5.5, 5.6, 5.7 compile error: 
Property 'import' in type 'Store<Q_RDF, Q_N3, OutQuad, InQuad>' is not assignable to the same property in base type 'Dataset<OutQuad, InQuad>'.
  Type '(stream: Stream<Q_RDF>) => EventEmitter<DefaultEventMap>' is not assignable to type '(stream: Stream<InQuad>) => Promise<this>'.
    Types of parameters 'stream' and 'stream' are incompatible.
      Type 'Stream<InQuad>' is not assignable to type 'Stream<Q_RDF>'.
        Type 'InQuad' is not assignable to type 'Q_RDF'.
          'InQuad' is assignable to the constraint of type 'Q_RDF', but 'Q_RDF' could be instantiated with a different subtype of constraint 'BaseQuad'  @definitelytyped/expect
  315:5  error  TypeScript@4.8, 4.9, 5.0 compile error: 
Property 'forEach' in type 'Store<Q_RDF, Q_N3, OutQuad, InQuad>' is not assignable to the same property in base type 'Dataset<OutQuad, InQuad>'.
  Type '(callback: QuadCallback<Q_N3>, subject: OTerm, predicate: OTerm, object: OTerm, graph: OTerm) => void' is not assignable to type '(callback: (quad: OutQuad, dataset: this) => void) => void'                                                                                                                                                                                                                                                                                                                               @definitelytyped/expect
  315:5  error  TypeScript@5.1, 5.2, 5.3, 5.4, 5.5, 5.6, 5.7 compile error: 
Property 'forEach' in type 'Store<Q_RDF, Q_N3, OutQuad, InQuad>' is not assignable to the same property in base type 'Dataset<OutQuad, InQuad>'.
  Type '(callback: QuadCallback<Q_N3>, subject: OTerm, predicate: OTerm, object: OTerm, graph: OTerm) => void' is not assignable to type '(callback: (quad: OutQuad, dataset: this) => void) => void'.
[...]

Issue #298, opened 2022-06-13
Migrate to Typescript


jeswr:
Would solve problems like #286 and #276 and anything else similar in the future.

RubenVerborgh:
Formal statement that this repository predates TypeScript 😆
But yes, great idea.

PR #636, opened 2026-07-02
fix: make import() return an awaitable promise for the store

jeswr:
Makes `Store#import()` awaitable: the returned stream is wrapped in a proxy that forwards `then`/`catch`/`finally` to a completion promise resolving to the store itself, satisfying the RDF/JS `Dataset.import(stream): Promise<Dataset>` signature while keeping the existing `Sink.import(stream): EventEmitter` behaviour intact for callers that never await.

Supersedes #503, reworked per the review notes there:

- **Breaking, so next major**: a thenable return value changes any `async` function that `return`s the result of `import()` — it now awaits stream completion and resolves to the store instead of the stream. Labelled `semver.major` accordingly.
- **No eager promise, no unhandled rejections**: the completion promise and its `end`/`error` listeners are only created upon first access of `then`/`catch`/`finally`. Non-awaiting Sink-style callers observe zero change: no listeners are added and an unlistened `error` event still throws loudly (an eagerly attached `error` listener — even guarded with a no-op `catch` — was verified to silently swallow such errors instead), and no rejection ever exists to go unhandled.
- **No EventEmitter/Store proxy as the resolution value**: the promise resolves to plain `this`, which is all `Promise<Store>` requires; the returned wrapper otherwise behaves as the stream.
- **Errored streams reject**: the already-completed check consults `stream.errored` before the ended/destroyed/non-readable checks, so a stream that has already errored (e.g. via `destroy(error)`) rejects instead of resolving as if it had ended.

Refs #298 #435 #503

cc @jeswr


src/N3Store.js
```diff
@@ -475,10 +475,46 @@ export default class N3Store {
...
+  // as a promise of the store (per the RDF/JS `Dataset.import` signature).
   import(stream) {
     stream.on('data', quad => { this.addQuad(quad); });
-    return stream;
+
+    // The completion promise is only created upon first access of
+    // `then`/`catch`/`finally`, such that callers who use the stream
+    // in the RDF/JS `Sink.import` style (without ever awaiting)
+    // observe no changes to the stream's listeners or error handling:
+    // an unhandled `error` event still throws, and no rejection occurs.
```
```suggestion
```

Issue #281, opened 2022-04-09
`matchPattern` method


jeswr:
In the case of patterns like `(?v, a, ?v)` these can be looked up much faster than matching against the blanks `(_, a, _)` by linking the keys within `_findInIndex`.

I propose a `matchPattern` method which is able to handle repeated variables of this nature when searching through the index.

Whilst they don't appear much in a raw query, these kinds of patterns can turn up when, for instance, performing reasoning; and as such I belive it is worthwhile having a faster way to look them up.

I've started to prototype this but haven't finished it yet.

jeswr:
More importantly - to enable much faster `SELECT DISTINCT` queries and reasoning over data sources it would be useful to either have as part of `matchPattern`; or even `match` itself, to specify that we only care about particular entries in a pattern. For instance running `match(false, null, false)` would essentially be a way of listing out all of the predicates in the graph. This can significantly reduce the number of quads that applications like Comunica would have to iterate over in some situations.

PR #637, opened 2026-07-02
feat: match Variables inside Quad pattern terms

### LINKED ISSUE #633: Index RDF 1.2 triple terms and match patterns inside reified quads

jeswr:
Adds structural matching for `Quad` pattern terms containing `Variable` components, on top of the component-id triple-term index of #635: `match`/`getQuads`/`countQuads`/`has` and friends can now answer `?r rdf:reifies <<( :s ?p ?o )>>`-style queries from the index instead of a full scan, in any pattern slot that can hold a triple term (subject, object, graph) and at any nesting depth.

Semantics (aligned with [rdf-stores](https://github.com/rubensworks/rdf-stores.js), so existing Comunica consumers understand them): concrete components must be term-equal; a `Variable` matches anything in its position; nested `Quad` components recurse; the same variable twice acts as independent wildcards (no join — that stays with #281); variable-free `Quad` patterns keep the exact single-id path; top-level `Variable` arguments stay non-matching; a pattern built with `DataFactory.quad(s, p, o)` matches default-graph triple terms, and a `Variable` graph component also matches API-only graph-component terms.

Implementation: wildcard-bearing slots expand into candidate term ids enumerated from the `_quadIds` index and loop the existing `_findInIndex`/`_countInIndex`/`indexMatch` machinery with the numeric ids as fixed keys; variable-free patterns and non-`Quad` slots take the previous code paths unchanged. Advertising `features.quotedTripleFiltering` to Comunica is deferred to a follow-up (it belongs with an integration test).

**Stacked on #635**, rebased directly onto current `main` without #601's cross-entity-index set-operation changes. The branch now contains the focused triple-term component-id index plus #637's structural matching, documentation, and array-pattern refactor.

Performance (M1, Node 25, interleaved A/B fresh-process runs vs the #635 tip 72c8ffd; workload per the #633 analysis, 100 k / 1 M quads): the wildcard query `<<( :s0 ?p ?o )>>` with 100 hits answers in **183 µs / 184 µs** — O(matches), size-independent — vs a 69 ms / 850 ms full-scan workaround (~375× / ~4 600×); the low-selectivity `<<( ?s :p3 ?o )>>` (N/16 hits) is materialization-bound at **15.6 ms / 168 ms** (~5× vs scan, ~3.5–4× faster than the design memo's prototype). Ingest, heap (byte-identical) and the variable-free exact-match paths are unchanged within noise (the variable-free paths are unchanged code). Full `npm test` (6 573 tests, 100 % coverage) and the RDF 1.2 syntax suites pass.

Open questions (from the design memo, for sign-off):

1. **Top-level `Variable` arguments**: kept non-matching (0 results, as today, pinned in a test). Aligning with rdf-stores (wildcards) would be more consistent but changes observable behavior of existing calls — separate decision.
2. **`forSubjects`/`forPredicates`/`forObjects`**: scope-cut to exact-only for `Quad` patterns in this PR (documented); the candidate loop can be extended there too if wanted.

Closes #633
Refs rdfjs/types#68

cc @jeswr


README.md
```diff
@@ -384,6 +384,20 @@ The store provides the following search methods
...
 
+In addition to `null` wildcards, a `Quad` pattern term passed to `match`, `getQuads`, `countQuads`, `has`, and friends may contain `Variable` components at any depth: concrete components must be equal, each `Variable` matches anything in its position independently (even when the same variable occurs twice), and nested `Quad` components recurse. This matches RDF 1.2 triple terms structurally, like `?r rdf:reifies <<( :s ?p ?o )>>` in SPARQL 1.2, and is aligned with [rdf-stores](https://github.com/rubensworks/rdf-stores.js) (see [#633](https://github.com/rdfjs/N3.js/issues/633)):
+
+```JavaScript
+// All quads reifying a statement about :s — ?r rdf:reifies <<( :s ?p ?o )>>
+const { namedNode, variable, quad } = N3.DataFactory;
+const reifications = store.getQuads(
+  null,
+  namedNode('http://www.w3.org/1999/02/22-rdf-syntax-ns#reifies'),
+  quad(namedNode('http://ex.org/s'), variable('p'), variable('o')),
```
jeswr: Since it is possible to _store_ variable terms in the store, we also do want to be able to actually match over variables; not just use variables to indiciate positions. Therefore I would suggest that we use arrays to indicate quad patterns that we want to match against and the above would be represented as `[namedNode('http://ex.org/s'), null, null, null]`

test/N3Store-test.js
jeswr, line 1: We should also make sure that this matching support deeply nested quoted triples. Please make sure there are tests for that if they are not already here.

jeswr:
As a separate PR it would make sense to introduce something like `matchPattern` support where a set of quads are provided which can include variable terms - representing a BGP pattern; and the set of variable bindings matched over the store are returned.

jeswr:
This PR appears to introduce cross graph logic - I assume because it builds upon a PR that has now been stood down. Can all of this logic please be removed.

jeswr:
Addressed. I rebuilt this branch directly on current `main`, retaining the focused triple-term component-id index and Quad-pattern matching while removing the inherited cross-entity-index `intersection`/`difference` helpers, remapping code, associated tests, and `N3StoreSetOps` benchmark. The full suite passes (6,637 tests, 100% coverage) and lint is clean. Current head: `5f98fdf`.

PR #638, opened 2026-07-02
feat: add a graphs option to control named-graph handling in the writer

### LINKED ISSUE #165: N-Quads are written if N-Triples are requested and graph is supplied

jeswr:
This adds a `graphs` option to `N3.Writer` that controls how quads in a named graph are serialized, implementing the design settled in #165 (`graphs: 'keep' | 'ignore' | 'error'`, default `'keep'`, per @RubenVerborgh's final comment there and his approval of #319):

- `'keep'` (default): current behaviour — graph names are written;
- `'ignore'`: the triple is written without its graph name (per the #165 discussion, this omits the graph name rather than dropping the quad);
- `'error'`: an error is raised for any quad in a non-default graph.

As in the original branch, the option applies uniformly in all formats, including N-Quads/TriG.

Supersedes #339 (its head branch lives on the upstream repo and cannot be pushed to; Laurens Rietveld's original commit is preserved via cherry-pick). On top of it, this branch fixes the callback firing twice — and the quad still being written — after an error in `_writeQuad`, keeps `quadToString` synchronous (it throws; `_writeQuadLine` routes the error to the callback), validates the option value in the constructor, and ports/extends the tests to the current jest style.

Note: #165 suggests flipping the default to `'error'`; that would be a breaking change and a candidate for the next major, so it is left out here.

Closes #165

cc @jeswr

### DIFF
--- a/README.md
+++ b/README.md
@@ -229,6 +229,12 @@ To write N-Triples (or N-Quads) instead, pass a `format` argument upon creation
 const writer1 = new N3.Writer({ format: 'N-Triples' });
 const writer2 = new N3.Writer({ format: 'application/trig' });
 ```
+To exclude the graph name from the serialization, pass a `graphs` argument upon creation:
+```JavaScript
+const writer3 = new N3.Writer({ graphs: 'keep' }); // Keeps the graph names (default)
+const writer4 = new N3.Writer({ graphs: 'ignore' }); // Ignores the graph name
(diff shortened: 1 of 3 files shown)


jeswr:
We should consider aligning this with the validation modes in https://github.com/rdfjs/N3.js/pull/634 and it being a validator that quads are piped through before the writing rather than done within the writer itself.

PR #639, opened 2026-07-02
chore(tests): run browser bundle tests in Chromium, Firefox and WebKit

### LINKED ISSUE #516: chore: Browser tests
We should add some browser tests so less manual validation / docs checking is required for PRs like #515

jeswr:
Adds a real-browser check for the shipped bundles: a standalone Playwright runner (`test/browser/run.mjs`) serves the repository over local HTTP, loads `test/browser/test-page.html` in Chromium, Firefox and WebKit, and verifies that both `browser/n3.min.js` (via `<script>` / `window.N3`) and `browser/n3.esm.min.js` (via `import()`) survive a parse → write → parse round trip including a language-tagged literal and an RDF 1.2 triple term `<<( )>>`. This is the check that would have answered #515's `queueMicrotask` question automatically. All six engine × bundle combinations pass locally (macOS, Playwright 1.61.1).

**Stacked on #608** (`deps/replace-browserify-esbuild`): the bundles under test are the esbuilt ones, so this branch is based on that PR's tip and its diff shows here until #608 lands.

Notes:

- The runner lives outside jest on purpose: the 100 % coverage gate measures `src/`, whereas this exercises build artifacts. `npm test` and its coverage are unchanged.
- CI gets a single non-matrix `browser` job (reuses the `build` artifact, `npx playwright install --with-deps`, runs the script), included in the `summary` required check; if it ever proves flaky it can be dropped from `summary.needs` without touching branch protection.
- `playwright` is dev-only (Node ≥ 18); jest never loads it, so the Node 18/20 CI floors are unaffected.
- `.gitignore`'s unanchored `browser` entry also ignored `test/browser/`; it is now anchored to the root build output (`/browser`).

Closes #516

cc @jeswr


jeswr (commented):
LGTM

jeswr:
conflicts need resolving

Issue #508, opened 2025-03-10
Notation3: Cannot parse Resource Paths used like an incomplete statement


tpluscode:
I use `eye` to process rules which use shorthand resource paths.

For example,

```turtle
"Some message"^log:trace .
```

This would be equivalent to 

```turtle
[] log:trace "Some message" .
```

n3.js fails with error message `Expected entity but got . on line 1.`

tpluscode:
Update: looks like resource paths are not supported at all?

jeswr:
Please see requested sanity check in #509

jeswr:
If anyone is looking to pick this up - I'm fairly sure paths *are* supported but buggy on some cases such as lists. There was work some work done to fix this in https://github.com/rdfjs/N3.js/pull/341, specifically https://github.com/rdfjs/N3.js/commit/be1bcd92059f02b4b9e03421be783844a178eb96

jimsmart:
> paths are supported but buggy on some cases such as lists.

Yeah, I seem to recall that currently N3Parser has paths working fine for IRIs/prefixed names, but - as you say - they're not currently implemented after lists... and also not after literals. (I have recently ported some of the N3.js code, so some of it is still fairly fresh in my memory)

This might be due to Notation3 spec changes since the original implementation of N3Parser's N3 support? Although I seem to recall previous specs were somewhat inexact in places, so it's difficult to be certain. It may just have been overlooked.

jeswr:
Confirmed still failing on 2.1.0 (`Expected entity but got .`). To pin down the current state: paths after IRIs and prefixed names do parse correctly; the parser just never looks for `!`/`^` after a literal (this issue), after a closing `)`, or after variables inside lists (the latter two are #348 — its example trips on the variable case first) — `_getPathReader` is only entered from the entity branches, matching @jimsmart's recollection. Since one fix covers all of these positions, I'm closing this as a duplicate of #348, the umbrella for complete resource-path support — the literal case belongs there too, so please follow along there.

PR #641, opened 2026-07-02
fix: support resource paths after literals, lists and variables in N3 mode

### LINKED ISSUE #348: Support resource path syntax
Defined in: https://w3c.github.io/N3/spec/#paths

Here is an [example](https://n3-editor.herokuapp.com/n3/editor/s/aSEPwg59) of a concise rule which drops the unnecessary intermediate variables using the `!` for resource paths: 

```turtle
@prefix string: <http://www.w3.org/2000/10/swap/string#> .
@prefix log: <http://www.w3.org/2000/10/swap/log#>.
@prefix : <https://example.com/> .

:this :badUrl <http://wrong.com/foo> .

{
  :this :badUrl ?badUrl .


  ?goodUrl log:uri ("http://good.stuff/" ( ?badUrl!log:uri "/(\\w+)$" )!string:scrape)!string:concatenation .
}
=>
{
  :this :goodUrl ?goodUrl .
} .
```

It [fails to parse](https://runkit.com/embed/nwpnuzga4z4r) and would be equivalent to

```turtle
@prefix string: <http://www.w3.org/2000/10/swap/string#> .
@prefix log: <http://www.w3.org/2000/10/swap/log#>.
@prefix : <https://example.com/> .

:this :badUrl <http://wrong.com/foo> .

{
  :this :badUrl ?badUrl .

  ?badUrl log:uri ?badUrlStr .
  ( ?badUrlStr "/(\\w+)$" ) string:scrape ?foo .
  ("http://good.stuff/" ?foo) string:concatenation ?gooUrlStr .
  
  ?goodUrl log:uri ?gooUrlStr .
}
=>
{
  :this :goodUrl ?goodUrl .
} .
```

jeswr:
In N3 mode, `!`/`^` resource paths were only recognised after IRIs and prefixed names in entity positions. This left several spec-valid path forms unparseable:

- paths starting with a literal — `"lit"!:p`, including datatyped, language-tagged and numeric/boolean literals, in subject, object and list-item positions;
- paths starting with a list — `(:a)!:p`, in subject, object and nested-list positions;
- paths after variables and blank node labels inside lists — `(?s!foaf:birthday)`, which is what the example in #348 trips over first;
- a path as a statement by itself — `"Some message"^log:trace .` (valid per `triples ::= subject predicateObjectList?`), the exact case of #508.

The parser now routes these through the existing path-reading machinery, reusing the `item` context so that list membership quads are deferred until the path has been read. The lexer additionally allows `!`/`^` to terminate blank node labels, numbers and booleans, mirroring `_prefixed`/`_variable`; Turtle/TriG still reject paths (now at the `!`/`^` token instead of inside the preceding token).

Test expectations are hand-derived from the N3 path semantics (`a!b` asserts `a b _:x` with the path denoting the fresh `_:x`; `a^b` the inverse) and were cross-checked against EYE for the literal, list, nested-list, empty-list, bare-statement and numeric/boolean/blank cases, including the full example from #348. The RDF 1.1/1.2 Turtle and TriG suites pass locally.

This ports the intent of `be1bcd9` from the abandoned `versions/2.0.0` branch onto `main`, but reimplements the list handling via the `item` context rather than a closure, and restricts the widened list-item condition to entity tokens so that `[`/`<<` items keep their current flow.

Overlap notes for in-flight PRs (this branch is based on `main`, not stacked):
- #613 touches `_completeSubjectLiteral`/`_readDirCode`; the pending-token drop after a subject literal is deliberately left in place here (only `!`/`^` are intercepted), and the new negative test was chosen so its error message is identical before and after #613.
- #617 and #628 touch `_readListItem`; small rebases expected for whichever lands later.
- #615 (blank node scoping): the new `(_:m!:p)` test encodes the current list-scoped label (`_:.m`) and would need its expectation updated by that PR.

Closes #348 (also resolves the case reported in #508, which was folded into it).

cc @jeswr


jeswr (commented):
LGTM

test/N3Parser-test.js
```diff
@@ -1305,6 +1305,24 @@ describe('Parser', () => {
...
+      shouldNotParse('<a> <b> true^<c>.',
+                     'Unexpected "^<c>." on line 1.'),
+    );
+
+    it(
+      'should not parse a ! path after a blank node',
+      shouldNotParse('<a> <b> _:m!<c>.',
+                     'Unexpected "!<c>." on line 1.'),
+    );
+
```
jeswr, line 1343: Please identify from which part of the N3 definition of paths this assumption can be drawn.

PR #642, opened 2026-07-03
perf(reasoner): remove per-match object/closure allocations in the inner loop

jeswr:
### What

Removes per-match allocations from the reasoner's hot path:

- `_evaluatePremise` allocated a fresh single-key `{ [value]: index[value] }` object at each of the three premise levels just to drive a `for-in` that ran exactly once when the variable was bound. Each level now branches directly: unbound → iterate the index keys; bound → a single index lookup. Match semantics are preserved exactly, including the level-2 bound case that runs once with no presence guard (leaves are stored as `null`) and the `Number(value)` key reconstruction.
- Conclusion emission no longer allocates a `forEach` callback plus a nested `() => cb(c)` closure per match; a reusable `_emit` method and a `cb(c)` signature on `_add` preserve the fixpoint signalling protocol.

Upstreams a June optimisation authored by @jeswr (OPT-16/OPT-43), verified against the June optimisation branch and re-measured on current `main`.

### Measured

Interleaved A/B fresh-process runs (alternating order per pair; median of in-process iterations, then median across pairs) of built `lib/` bundles on an otherwise idle macOS/arm64 machine. Baseline is `main` (695f908).

| workload | pairs | main (ms) | this PR (ms) | delta |
|---|---|---|---|---|
| reason: 10^4-deep subclass chain + subclass rule | 7 | 30.67 | 20.24 | **−34.0%** |
| same workload, independent 5-pair run | 5 | 22.65 | 16.49 | **−27.2%** |
| reason: full RDFS ruleset over timbl+foaf (1830 quads out) | 5 | 26.64 | 12.15 | **−54.4%** |
| store add/query (add-grid, add-parsed, add-duplicates, s??, ?p?, ??o, sp?, s?o, full-scan) | 5–7 | — | — | −5.0%…+2.6%, within noise |

The built output differs only in `lib/N3Reasoner.js`, so store paths are untouched by construction; the one store workload that first showed +8.6% re-measured at −3.2% over 7 pairs (noise).

### Correctness

Beyond the full jest suite (100% coverage), the reasoner was run differentially old-vs-new on three rulesets — the 10^4-deep taxonomy (20 001 quads), the full RDFS ruleset over timbl+foaf (1 830 quads), and a multi-premise ruleset (subclass transitivity over a branching class tree plus a two-premise `knows` join, 1 398 quads) — asserting byte-identical sorted inferred-quad sets in every case. A new test covers the previously unhit level-2-bound, non-last-premise branch.

cc @jeswr


src/N3Reasoner.js
```diff
@@ -18,44 +18,91 @@ export default class N3Reasoner {
...
     // Only add to the remaining indexes if there is not already a value in the index
     if (!this._store._addToIndex(graphItem.subjects,   subject,   predicate, object)) return;
     this._store._addToIndex(graphItem.predicates, predicate, object,    subject);
     this._store._addToIndex(graphItem.objects,    object,    subject,   predicate);
-    cb();
+    cb(c);
   }
 
+  // OPT-43: invoked once per fully-bound match; pulled out of _evaluatePremise so the
+  // hot loop no longer allocates a fresh callback closure per match.
```
jeswr: These comments should be understandable standalone. Don't include OPT codes here.

src/N3Reasoner.js
```diff
@@ -18,44 +18,91 @@ export default class N3Reasoner {
...
+
+  // OPT-16: at each of the three premise levels, when the variable is bound we do a
+  // direct index lookup instead of allocating a single-key `{ [value]: index[value] }`
+  // object purely to drive a for-in that only ever runs once. The unbound case still
+  // iterates every key of the index. for-in yields string keys and `index[value]`
+  // coerces a numeric `value` to string, so the lookups are identical; the
+  // `val.value = Number(value)` reconstruction (unbound only) and the truthiness guards
+  // are preserved exactly.
+  // OPT-43: conclusion emission is delegated to _emit (a method, not a fresh closure per
+  // match), preserving the cb(c) signalling protocol that drives the fixpoint loop.
```
jeswr: Ditto with the opt codes

src/N3Reasoner.js
jeswr: Done in b7121ea — removed the OPT codes from all comments in the diff; they now read standalone.

src/N3Reasoner.js
jeswr: Done in b7121ea.

PR #643, opened 2026-07-03
perf(parser): reduce per-quad allocation (no throwaway literal, no per-literal .bind, lean contexts)

jeswr:
Reduces per-quad allocation in the parser with three small changes, at no wall-time cost:

- **No throwaway literal**: `_completeLiteral` built a plain-string `Literal` for every literal and discarded it whenever a datatype or language tag followed; the default literal is now only built in the fall-through branch.
- **No per-literal `.bind`**: the language+direction completion path allocated a fresh bound function per language-tagged literal; deferral is now a boolean plus pending state on the instance, read back by a fixed `_readDirCode(token)` method.
- **Lean non-N3 contexts**: `_saveContext` pushed a full N3 context object even for Turtle/TriG/N-Triples/N-Quads, where the N3-only fields are unused; non-N3 modes now push a five-field context.

Output is byte-identical to `main` (SHA-256 over the quad streams of the five benchmark documents). Allocation churn (sampling heap profiler, bytes allocated per quad parsed; a second run reproduced every row within ±1.2%):

| document | main | this PR |
|---|---|---|
| Turtle, 500k | 1641 | 1565 (−4.6%) |
| TriG, 500k / 10 graphs | 1701 | 1655 (−2.7%) |
| N-Triples, 500k | 1759 | 1681 (−4.4%) |
| N3, formulae + lists, 780k | 1107 | 1105 (−0.2%; N3 keeps the full context shape by design) |
| Turtle 1.2, dir-lang + annotations, 600k | 1523 | 1400 (−8.1%) |

Wall time is neutral within the machine's ±4% noise band (interleaved A/B, seven fresh-process pairs per format): this is an allocation/GC-pressure reduction, not a speed-up claim.

Deliberately excluded: a per-base `Map<rawIRI, resolved>` cache in `_resolveIRI` measured a **+16.1% wall-time regression** on fragment-relative documents because resolving `#x` is cheaper than a `Map.get`.

Rebased onto current `main`, preserving the newer empty-formula and graph-term parser behavior. The full 6,725-test suite passes with 100% measured coverage; lint and the Node build also pass.

cc @jeswr

### DIFF
--- a/src/N3Parser.js
+++ b/src/N3Parser.js
@@ -78,31 +78,30 @@ export default class N3Parser {
   // ### `_saveContext` stores the current parsing context
   // when entering a new scope (list, blank node, formula)
   _saveContext(type, graph, subject, predicate, object) {
-    const n3Mode = this._n3Mode;
+    // Only N3 contexts need the extra parser state.
+    if (!this._n3Mode) {
+      this._contextStack.push({ type, subject, predicate, object, graph });
+      return;
+    }
     this._contextStack.push({
       type,
       subject, predicate, object, graph,
-      inverse: n3Mode ? this._inversePredicate : false,
-      blankPrefix: n3Mode ? this._prefixes._ : '',
-      quantified: n3Mode ? this._quantified : null,
-      emptyFormula: n3Mode ? this._emptyFormula : false,
+      inverse: this._inversePredicate,
+      blankPrefix: this._prefixes._,
+      quantified: this._quantified,
+      emptyFormula: this._emptyFormula,
     });
-    // The settings below only apply to N3 streams
-    if (n3Mode) {
-      // Every new scope resets the predicate direction
-      this._inversePredicate = false;
-      // In N3, blank nodes are scoped to a formula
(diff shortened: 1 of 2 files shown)


src/N3Parser.js
```diff
@@ -75,24 +75,30 @@ export default class N3Parser {
   // ### `_saveContext` stores the current parsing context
   // when entering a new scope (list, blank node, formula)
   _saveContext(type, graph, subject, predicate, object) {
-    const n3Mode = this._n3Mode;
+    // The inverse/blankPrefix/quantified fields only apply to N3 streams.
+    // For non-N3 modes (Turtle/TriG/N-Triples/N-Quads) they are always
+    // false/''/null, so push a lean 5-field context object instead (OPT-27).
+    // A parser instance has a fixed mode, so only one shape is ever pushed
```
jeswr: OPT-27 doesn't mean much in this context

jeswr: Done in d12f2c9 — dropped the OPT references (here and one further down in _completeLiteral); the comments now stand alone.

PR #645, opened 2026-07-03
perf: cache literal term getters

jeswr:
Reworked per review: the single-pass writer encoder is reverted (its edge-case tests are kept) and `Literal`'s `value`/`language`/`direction`/`datatype` getters now memoise on first access — each stores its result as a non-enumerable, non-writable own property that shadows the prototype getter, so repeated reads stop rescanning `id` and `datatype` returns a stable `NamedNode` per literal instead of a fresh one per access (nothing in the code base, test suite, or RDF/JS requires freshness; `equals`/`termToId`/`termFromId` operate on `id`, unchanged).

Design notes:
- **Lazy** — the constructor is untouched, so parsing pays nothing (parse of 1M literal triples: retained memory identical at 341.6 B/quad, CPU within noise).
- **Non-enumerable is load-bearing** — `_value`-style backing slots are visible to `deepStrictEqual`/`toEqual`/spreads, so an accessed literal would stop deep-equalling a fresh one; own non-enumerable data properties are invisible to all of these and instances stay in fast-properties mode.
- **Frozen/sealed literals** skip caching (`Object.isExtensible` guard) and keep recomputing as before.

CPU ratios vs main (interleaved main↔branch fresh-process pairs × 5 reps, process-CPU medians; Apple M1, Node 25):

| workload | getter caching (this PR) | single-pass encoder (reverted) |
|---|---|---|
| writer, literal-heavy mix, 1M → string | **0.820** | 0.877 |
| writer, lists, 200k → string | **0.782** | 0.776 |
| parse 1M literal triples, no getter access | 1.019 (noise); memory identical | untouched |
| read value + datatype ×2 per literal, 1M | **2.11** | untouched |

The last row is the honest cost: `Object.defineProperty` runs ~250 ns per cached getter, so literals whose getters are read only once or twice come out slower (breakeven ≈10 repeated reads per getter for short literals; far earlier for long ids, which is why the writer wins). A `_cache`-object variant (one `defineProperty` per literal) measured no better; enumerable-slot variants are excluded by the deep-equality argument.

Byte identity vs main verified over eight generated 20k-quad corpora plus ~60 adversarial terms through `Writer`→string, `quadsToString`, and `StreamWriter`. Tests: a new `Literal` caching suite (repeated-access identity, stable `datatype`, deep-equality and `equals` stability after access, non-enumerability, frozen instances) plus the kept writer edge-case tests; coverage stays at 100%.

cc @jeswr


jeswr:
This is not a sensible approach. Instead lets cache the getters on literals

jeswr:
Reworked as requested: the single-pass writer change is reverted (edge-case tests kept, ec61276) and the `Literal` getters now memoise on first access as non-enumerable own properties — 290efd8. Lazy, so parse pays nothing, and cached literals stay deep-equal to fresh ones; the honest trade is that literals whose getters are read only once or twice get slower. Full numbers, byte-identity, and the excluded alternatives are in the updated description — your call.

jeswr:
Could you do an updated performance assessment where rather than doing `Object.defineProperty` and completely overriding the getter method - we have e.g., `_value` and `_language` which are set to null and then used to store the internal cache value and the firsst piece of logic in the getter does `if (this._value) return this._value ...`

PR #646, opened 2026-07-03
perf: coalesce stream writer output into 16 KiB chunks

jeswr:
`N3StreamWriter` currently pushes every serialized fragment as its own chunk. Each push pays the stream machinery and downstream string-to-buffer conversion costs, so this PR coalesces fragments in the writer shim and flushes full chunks at 16 KiB. `end()` always flushes the tail.

Partial chunks flush after a bounded 20 ms delay. The timer starts when the buffer becomes non-empty rather than debouncing each fragment, so a steady sub-threshold stream cannot postpone output indefinitely. The existing `flushDelay` option can override the latency bound. The timer is unreferenced where supported and is cancelled on size flush, `end()`, or `destroy()`.

### Default selection

The 16 KiB threshold matches the readable high-water mark used by this repository's `readable-stream` 4.7 dependency. Profiling on an Apple M1 with Node 25 serialized 200,000 quads (13.38 MB) five times per threshold in rotating order:

| Threshold | Median CPU | Emitted chunks |
| --- | ---: | ---: |
| Uncoalesced | 202.7 ms | 200,001 |
| 4 KiB | 167.3 ms | 3,224 |
| 8 KiB | 165.8 ms | 1,625 |
| 16 KiB | 165.6 ms | 816 |
| 32 KiB | 166.7 ms | 409 |
| 64 KiB | 175.5 ms | 205 |

16 KiB was the fastest measured threshold and avoids buffering four times the stream's own high-water mark.

For paced input, 20 ms reduced 120 emissions at a 1 ms arrival cadence to 7 with a measured maximum latency of 20.8 ms. At 5 ms and 10 ms cadences it reduced 60 emissions to 15 and 30 respectively. The previous 100 ms value reached 100.8 ms measured latency; the new default preserves useful coalescing while cutting that bound by five.

### Observable behavior and verification

The byte stream is unchanged, but consumers observe fewer, larger chunks. Error paths flush buffered output before emitting the error. Tests cover byte identity, chunk thresholds, timer firing and re-arming, custom delay, timer cancellation, browser timer compatibility, and serialization/input errors.

Local verification: 6,626 tests, lint, and the Node build pass with 100% measured source coverage.

### DIFF
--- a/src/N3StreamWriter.js
+++ b/src/N3StreamWriter.js
@@ -5,6 +5,9 @@ import N3Writer from './N3Writer';
 // Minimum size of coalesced output chunks
 const MIN_CHUNK_SIZE = 65536;
 
+// Default maximum time (in milliseconds) that output stays coalesced
+const DEFAULT_FLUSH_DELAY = 100;
+
 // ## Constructor
 export default class N3StreamWriter extends Transform {
   constructor(options) {
@@ -15,12 +18,23 @@ export default class N3StreamWriter extends Transform {
     // conversion run once per chunk rather than once per serialized fragment
     this._buffer = '';
 
+    // Slow quad sources (e.g., remote query results arriving in real time)
+    // could otherwise leave output stuck below the minimum chunk size,
+    // so a timer bounds how long output stays buffered.
+    // It is armed when the buffer becomes non-empty and cleared on flush,
+    // keeping it out of the per-fragment hot path.
+    this._flushTimer = null;
+    this._flushDelay = options && options.flushDelay !== undefined ?
+      options.flushDelay : DEFAULT_FLUSH_DELAY;
+
     // Set up writer with a dummy stream object
     const writer = this._writer = new N3Writer({
       write: (chunk, encoding, callback) => {
         this._buffer += chunk;
         if (this._buffer.length >= MIN_CHUNK_SIZE)
           this._pushBuffer();
+        else if (this._flushTimer === null)
+          this._armFlushTimer();
         callback && callback();
       },
       end: callback => { this._pushBuffer(); this.push(null); callback && callback(); },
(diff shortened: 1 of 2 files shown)


jeswr:
There may be unintended consequences for tools like `comunica` where results are streamed in from a remote source very slowly and they need to be printed in real-time. For use cases like that one we will want to detect a sustained pause. E.g. of 0.1 seconds from the last chunk emitted, and trigger an emission even if we are not at 64kb.

jeswr:
Good point — implemented in 3615fa5: a timer bounds how long output stays buffered, flushing `flushDelay` ms after the buffer becomes non-empty (default 100 ms, constructor option) unless the 64 KB flush got there first — so any ≥100 ms pause triggers an emission, and even a steady sub-64 KB trickle that never pauses is delivered within 100 ms. Coalescing win preserved and byte identity re-verified with paced runs where the timer fires; details in the updated description.

jeswr:
The only thing I am still unsure about here are the defaults of 64kb and 100ms. Specifically, it feels like 100ms may be unnecessarily long to achieve the desired effect and 64kb. Could you do some research and/or profiling to work out what good defaults are here.

jeswr:
Profiled both defaults and updated them in `149bf41`:

- **16 KiB chunk threshold:** this matches `readable-stream` 4.7's 16 KiB readable high-water mark and had the lowest median CPU time in a five-run rotating matrix (165.6 ms for 200k quads, versus 175.5 ms at 64 KiB and 202.7 ms uncoalesced).
- **20 ms flush delay:** with quads arriving every 1 ms, this reduced 120 emissions to 7 while keeping measured maximum latency to 20.8 ms. At 5 ms and 10 ms arrival cadences it reduced 60 emissions to 15 and 30. The old default measured about 100.8 ms maximum latency.

The PR description now includes the profiling table and methodology. I also rebased onto current `main`; all 6,626 tests, lint, and the Node build pass.

PR #647, opened 2026-07-03
perf: track index entry counts for constant-time removal cleanup

jeswr:
`removeQuad` checks whether the graph became empty with a `for`…`in` over **all** subjects in the graph, and `_removeFromIndex` uses the same idiom on the two inner index levels. Large index levels are dictionary-mode objects, so `for`…`in` collects (and sorts) the full key set before the first iteration — the early `return` saves nothing, making every removal O(#subjects in the graph). `deleteMatches`, `removeMatches`, `deleteGraph`, and reasoner churn all sit on this path. (The V8-side cost is reported upstream: https://issues.chromium.org/issues/530857250.)

This PR keeps an entry counter on every index node under a `Symbol` key (invisible to `for`…`in` and `Object.keys`), so empty-level cleanup and the graph-emptiness check are O(1). The invariant: every function that builds index nodes maintains the counter — `_addToIndex`, `merge`, `intersect`, `difference`, and `indexMatch` — which covers add/remove/`removeMatches`/`deleteMatches`, stores produced by `addAll`/`union`/`difference`/`intersection`, mutation of materialized `match()` views, and `N3Reasoner`.

Measured (fresh process per measurement, interleaved A/B vs main, medians of 5 rounds, checksum-gated; Apple M1, Node 25):

| workload | main | this PR |
|---|---:|---:|
| removeQuad @100k / @1M | 1 301 / 15 322 µs/op | 3.0 / 14.4 µs/op (~430× / ~1 060×) |
| deleteMatches(s,·,·) @100k / @1M | 12 749 / 153 508 µs/op | 55.7 / 202 µs/op (~230× / ~760×) |
| addQuad fresh load @100k / @1M | 5.05 / 5.44 µs/op | 3.84 / 5.75 µs/op (no regression; ±5 % noise band) |
| memory @1M | 2 457.0 B/quad | 2 449.3 B/quad (−0.3 %) |

Correctness: beyond the existing suite, a randomized differential test (3 fixed seeds × 400 mixed operations — add/remove/deleteMatches/addAll, set operations, materialized-view mutation, full drain) against a naive `Set` oracle, plus an audit that every index node's counter equals its actual entry count.

In-flight PRs: #601 touches the same `merge`/`intersect`/`difference` helpers — whichever lands second applies the counter discipline to the new helper bodies (mechanical; the invariant is documented at the `SIZE` declaration). #635 (term-id encoding) has no structural overlap; #598's delta-based view removals inherit the win.

cc @jeswr

### DIFF
--- a/src/N3Store.js
+++ b/src/N3Store.js
@@ -7,12 +7,27 @@ import N3Writer from './N3Writer';
 
 const ITERATOR = Symbol('iter');
 
-function merge(target, source, depth = 4) {
-  if (depth === 0)
-    return Object.assign(target, source);
+// `SIZE` tracks the number of entries in an index node under a Symbol key,
+// which is invisible to `for...in` loops and `Object.keys`.
+// It makes empty-layer cleanup during removal O(1) instead of O(entries).
+// Every function that creates index nodes
+// (`_addToIndex`, `merge`, `intersect`, `difference`, and `indexMatch`)
+// must maintain it on the three levels below a graph.
+const SIZE = Symbol('size');
 
-  for (const key in source)
-    target[key] = merge(target[key] || Object.create(null), source[key], depth - 1);
+function merge(target, source, depth = 4) {
+  let size = target[SIZE] || 0;
+  for (const key in source) {
+    if (!(key in target)) {
+      size++;
+      target[key] = depth === 0 ? null : merge(Object.create(null), source[key], depth - 1);
+    }
+    else if (depth !== 0)
+      target[key] = merge(target[key], source[key], depth - 1);
+  }
+  // Depth 2 is the level of the `subjects`, `predicates`, and `objects` indexes.
+  if (depth <= 2)
+    target[SIZE] = size;
 
   return target;
 }
@@ -25,7 +40,7 @@ function merge(target, source, depth = 4) {
  * *not* be set as the value for an index.
  */
 function intersect(s1, s2, depth = 4) {
-  let target = false;
+  let target = false, size = 0;
 
   for (const key in s1) {
     if (key in s2) {
@@ -33,6 +48,7 @@ function intersect(s1, s2, depth = 4) {
       if (intersection !== false) {
         target = target || Object.create(null);
         target[key] = intersection;
+        size++;
       }
       // Depth 3 is the 'subjects', 'predicates' and 'objects' keys.
       // If the 'subjects' index is empty, so will the 'predicates' and 'objects' index.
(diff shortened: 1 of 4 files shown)


jeswr:
Is there not some way of using Object.size or something like that to check the size of the object rather than needing to track this ourselves? Secondly, this seems like an underlying bug in the V8 engine used to evaluate javascript in node and chrome - I would not have thought that early stopping for loops should have a performance penalty which scales with the size of the object. Could you investigate this upstream, identify the bug - and open a pull request to fix it.

jeswr:
Result of investigation https://issues.chromium.org/issues/530857250

test/N3Store-test.js
jeswr, line 1: Something similar to this recently landed on main. Please ensure there is no duplicate logic.

src/N3Store.js
```diff
@@ -7,12 +7,27 @@ import N3Writer from './N3Writer';
...
-function merge(target, source, depth = 4) {
-  if (depth === 0)
-    return Object.assign(target, source);
+// `SIZE` tracks the number of entries in an index node under a Symbol key,
+// which is invisible to `for...in` loops and `Object.keys`.
+// It makes empty-layer cleanup during removal O(1) instead of O(entries).
+// Every function that creates index nodes
+// (`_addToIndex`, `merge`, `intersect`, `difference`, and `indexMatch`)
+// must maintain it on the three levels below a graph.
+const SIZE = Symbol('size');
```
jeswr, line 9: Can the introduction of size be used to optimise any other functionality (e.g. the `SIZE` call on `Store`s - and selecting which stores & indices to make the left and right when doing operations like merge/intersect/difference etc,

src/N3Store.js
```diff
@@ -7,12 +7,27 @@ import N3Writer from './N3Writer';
 
 const ITERATOR = Symbol('iter');
```
jeswr: We should open a ticket to take this logic back out when most popular javascript engines properly optimise this type of behavior.

jeswr: Opened #684 to track removing the counters once supported JavaScript engines optimize the underlying empty-index check: https://github.com/rdfjs/N3.js/issues/684

test/N3Store-test.js
jeswr: Rebased onto current main and extracted the seeded makeRng and pick helpers introduced by #652 into test/util.js. The store mutation suite now reuses that shared logic and retains only its store-specific oracle and counter-invariant checks.

src/N3Store.js
jeswr: Applied the counters to cached size recomputation and wildcard quad counting, and intersection now iterates the smaller counted index at each indexed level. Merge mutates its left operand and difference has ordered semantics, so swapping those operands would not be safe.

PR #648, opened 2026-07-03
perf: answer fully bound has() with a direct index probe

jeswr:
A fully bound `has()` — every DatasetCore-style `has(quad)` — currently runs the whole `readQuads` generator machinery (bound-graph object literal, `_findInIndex`, term and quad materialisation, iterator records) to produce one boolean, allocating ~2.5 KB per call. This adds a direct index probe for the fully bound case; patterns with unbound components keep the generator path.

Semantics are unchanged:
- a term (or Quad term) that does not resolve to an interned id ⇒ `false`, as on main;
- the default graph resolves via `isDefaultGraph` ⇒ id 1; string ids keep working; an unbound graph stays a wildcard pattern through `readQuads`;
- RDF-star: fully bound quoted-quad terms go through the exact composite-id lookup, matching main's "unresolvable ⇒ false" behaviour;
- `match()` views (`DatasetCore.has`) resolve through the materialised filtered store and take the probe.

Measured — fresh process per measurement, interleaved A/B vs `main`, medians of 5 rounds, checksum-gated (Apple M1, Node v25, load < 8; realistic-shape store; each probe includes constructing the argument quad, so the pure `has()` delta is larger):

| workload | main | this PR | speedup |
|---|---:|---:|---:|
| has (hit) @100k | 5.42 µs/op | 0.87 µs/op | ~6.2× |
| has (hit) @1M | 6.71 µs/op | 1.18 µs/op | ~5.7× |
| has (miss, interned combination) @100k | 2.98 µs/op | 0.51 µs/op | ~5.8× |
| has (miss, interned combination) @1M | 3.60 µs/op | 0.85 µs/op | ~4.2× |
| allocation per hit | ~2 460 B | ~40 B | |
| addQuad fresh load | | | unchanged (1.01× / 0.99×) |

`contains`, `equals`, and the foreign-index set-operation fallbacks call `has` per quad and inherit the win.

Note for #637: Quad patterns containing Variables must not take the probe; on main both paths already answer `false` (the composite id does not resolve), and this PR preserves exactly that. Once #637's pattern classifier lands, the probe guard should defer wildcard-Quad patterns to it — flagged here for the rebase.

cc @jeswr


jeswr (commented):
Is it possible to refactor the code such that we get a similar performance improvement without the same impact on readability.

PR #652, opened 2026-07-04
test: add deterministic fuzz / round-trip property tests

jeswr:
Adds `test/Fuzz-test.js` with seeded (LCG) property tests, so the corpus is generated but every CI run is identical — no flaky or irreproducible failures. Each iteration is reduced to a boolean so the jest assertions stay unconditional:

- 1000 Turtle + 500 N3-mode arbitrary inputs: parsing either throws a catchable `Error` or yields structurally well-formed quads (RDF/JS term types, string values, `NamedNode` literal datatypes).
- 300 flat graphs whose literal values draw from an adversarial character pool (quotes, delimiters, backslashes, controls, astral): write → parse round-trips isomorphically.
- 150 graphs containing quoted triples (simple term values): write → parse round-trips isomorphically.

The suite runs in under a second and generates no on-disk data. All 6508 tests pass; coverage stays at 100%.

cc @jeswr

### DIFF
--- a/test/Fuzz-test.js
+++ b/test/Fuzz-test.js
@@ -1,0 +1,151 @@
+// Deterministic fuzz / round-trip property tests.
+//
+// These assert library-wide invariants rather than specific outputs:
+//  1. Parsing ARBITRARY input never crashes uncatchably — it either returns
+//     structurally well-formed quads or throws a catchable Error.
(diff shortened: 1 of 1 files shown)


jeswr (commented):
LGTM

PR #653, opened 2026-07-04
fix(datafactory): type unsafe-range numbers as xsd:double, not xsd:integer

jeswr:
`DataFactory.literal()` auto-types any `Number.isInteger` value as `xsd:integer`, which mis-types two ranges (reproduced on main):

- `literal(1e21)` → `"1e+21"^^xsd:integer` — an invalid integer lexical (JS stringifies |n| ≥ 1e21 in exponential form)
- `literal(2 ** 53 + 1)` → `"9007199254740992"^^xsd:integer` — the IEEE 754 value silently lost the exact integer, yet the datatype still claims exactness

Fixed by tagging `xsd:integer` only for `Number.isSafeInteger` values; everything else becomes `xsd:double`, whose value space is exactly a JS number. After: `literal(1e21)` → `"1e+21"^^xsd:double`.

Deliberate consequence to weigh: exactly-representable-but-unsafe integers such as `1e20` also move from `xsd:integer` to `xsd:double` (still a valid lexical) — beyond 2^53 neighbouring integers collapse onto the same double, so an integer typing can't be trusted. Happy to narrow the boundary to only the exponential-form range (|n| ≥ 1e21) if the safe-integer cut is considered too strict.

GIGO note: the input is a perfectly valid JS number — the wrong datatype comes from N3's own inference, so this is plain correctness, not validation.

cc @jeswr

### DIFF
--- a/src/N3DataFactory.js
+++ b/src/N3DataFactory.js
@@ -389,7 +389,13 @@ function literal(value, languageOrDataType) {
     // Convert an integer or double
     else if (typeof value === 'number') {
       if (Number.isFinite(value))
-        datatype = Number.isInteger(value) ? xsd.integer : xsd.double;
+        // Tag as xsd:integer only when the number provably denotes that exact
+        // integer. Beyond 2^53, a JS number can no longer represent every
+        // integer (9007199254740993 silently becomes ...992), and from 1e21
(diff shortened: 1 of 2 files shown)


jeswr:
Not sure on this one since we are converting _to_ xsd integers which are unbounded

PR #654, opened 2026-07-04
fix: guard termFromId JSON parsing of quoted-triple ids

jeswr:
`termFromId` parses a `[`-prefixed id (a quoted triple) with `JSON.parse`. A malformed or hostile id — invalid JSON, or JSON that isn't a 3-or-4-element array — threw or produced a malformed term. This wraps the parse in try/catch and validates the shape, falling back to `factory.namedNode(id)` for anything that isn't a well-formed quoted-triple id.

Tests: focused cases for invalid/malformed `[`-prefixed ids.

The `extractLists` null-prototype hardening that was originally bundled here is split into #668, per review.

### DIFF
--- a/src/N3DataFactory.js
+++ b/src/N3DataFactory.js
@@ -244,9 +244,21 @@ export function termFromId(id, factory, nested) {
     }
     return factory.literal(id.substr(1, endPos - 1),
             languageOrDatatype);
-  case '[':
-    id = JSON.parse(id);
+  case '[': {
+    // A `[`-prefixed id encodes a quoted triple as a JSON array of
+    // 3 or 4 term ids; ids that do not parse to that shape are named nodes
+    let parsed;
+    try {
+      parsed = JSON.parse(id);
+    }
+    catch (e) {
+      return factory.namedNode(id);
+    }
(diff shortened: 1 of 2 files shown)


src/N3Store.js
```diff
@@ -759,7 +759,10 @@ export default class N3Store {
   // ### `extractLists` finds and removes all list triples
   // and returns the items per list.
   extractLists({ remove = false, ignoreErrors = false } = {}) {
-    const lists = {}; // has scalar keys so could be a simple Object
+    // Null-prototype map: keys are the list heads' term values, so a plain `{}`
+    // would let a head value such as `toString` or `constructor` collide with
+    // the inherited members of the returned object (CWE-1321).
+    const lists = Object.create(null);
```
jeswr: This should be in a separate PR from the JSON parsing

jeswr:
Addressed: reduced to the termFromId JSON-parse guard; the extractLists null-prototype hardening is split into #668, and the title is updated to match.

jeswr:
We should not proceed with this PR - and just implement an RDF 1.2 index instead.

PR #655, opened 2026-07-04
fix(parser): bound input-derived content in error messages

jeswr:
Several parser error messages interpolate an input-derived term id, e.g. `Expected punctuation to follow "<IRI>"`, so the error string grows with the offending token (CWE-209 / CWE-770). Reproduced on main: a 3 MB IRI produced a 3,000,053-char `error.message`.

`_error` now caps the message at 200 chars with a `…` marker; the untruncated token remains available on the existing `err.context.token`. The lexer-side `Unexpected "…"` excerpt is a separate path and is left untouched here.

GIGO note: plain robustness of N3's own error construction — no change to what is accepted or rejected.

cc @jeswr


src/N3Parser.js
```diff
@@ -1083,6 +1083,12 @@ export default class N3Parser {
 
   // ### `_error` emits an error message through the callback
   _error(message, token) {
+    // Some messages interpolate input-derived content (e.g. a multi-MB IRI or
+    // literal), so bound the message to keep a single crafted token from
+    // producing an unbounded error string (CWE-209 / CWE-770). The full token
```
jeswr: Not necessary to reference CWE's in comments. Otherwise LGTM.

PR #656, opened 2026-07-04
fix(lexer): anchor the dircode regex

jeswr:
Fixes unanchored direction-code matching in the lexer.

The direction-code regex allowed the rtl alternative to match anywhere in the remaining input. As a result, an invalid direction followed later by rtl could be consumed as a valid direction code instead of producing an error. The regex now anchors both alternatives, and regression coverage verifies the invalid input is rejected.

The literal-scan cache reset that appeared in the initial draft was removed in commit 73b8dc7 following author review. Lexer reuse and the state that would need resetting are a separate concern and are not part of this PR.

### DIFF
--- a/src/N3Lexer.js
+++ b/src/N3Lexer.js
@@ -43,7 +43,9 @@ export default class N3Lexer {
     this._simpleQuotedString = /^"([^"\\\r\n]*)"(?=[^"])/; // string without escape sequences
     this._simpleApostropheString = /^'([^'\\\r\n]*)'(?=[^'])/;
     this._langcode = /^@([a-z]+(?:-[a-z0-9]+)*)(?=[^a-z0-9])/i;
-    this._dircode = /^--(ltr)|(rtl)/;
+    // Both directions require the `--` prefix; the unanchored form
+    // `/^--(ltr)|(rtl)/` matched `rtl` anywhere in the remaining input
+    this._dircode = /^--(?:(ltr)|(rtl))/;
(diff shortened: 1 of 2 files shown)


src/N3Lexer.js
```diff
@@ -43,7 +43,9 @@ export default class N3Lexer {
     this._simpleQuotedString = /^"([^"\\\r\n]*)"(?=[^"])/; // string without escape sequences
     this._simpleApostropheString = /^'([^'\\\r\n]*)'(?=[^'])/;
     this._langcode = /^@([a-z]+(?:-[a-z0-9]+)*)(?=[^a-z0-9])/i;
-    this._dircode = /^--(ltr)|(rtl)/;
+    // Both directions require the `--` prefix; the unanchored form
+    // `/^--(ltr)|(rtl)/` matched `rtl` anywhere in the remaining input
```
jeswr: You don't need to reference old versions of the code in comments

src/N3Lexer.js
```diff
@@ -511,6 +513,9 @@ export default class N3Lexer {
   // The input can be a string or a stream.
   tokenize(input, callback) {
     this._line = 1;
+    // Reset the long-literal scan cache, so a reused lexer
+    // cannot carry a stale offset from a previous document
+    this._literalClosingPos = 0;
```
jeswr: I don't see how this is related to the rest of the changes in the PR. Further, as I understand the lexer is not designed to be re-used; so would there not need to be more things reset like the context. Split this out into a separate PR.

src/N3Lexer.js
```diff
@@ -44,7 +44,7 @@ export default class N3Lexer {
     this._simpleQuotedString = /^"([^"\\\r\n]*)"(?=[^"])/; // string without escape sequences
     this._simpleApostropheString = /^'([^'\\\r\n]*)'(?=[^'])/;
     this._langcode = /^@([a-z]+(?:-[a-z0-9]+)*)(?=[^a-z0-9])/i;
-    this._dircode = /^--(ltr)|(rtl)/;
+    this._dircode = /^--(?:(ltr)|(rtl))/;
```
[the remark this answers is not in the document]
jeswr: Addressed by correcting the PR scope and metadata. The literal-scan cache change was intentionally removed in 73b8dc7 after the earlier author review requested that lexer-reuse behavior be handled separately. This PR now only claims and tests the direction-code regex fix; the title and description no longer state that the cache reset is included.

PR #657, opened 2026-07-04
feat(reasoner): opt-in maxDerivations and maxPremiseDepth budgets

jeswr:
Naive forward chaining is super-linear: a transitive-closure rule over a chain of n edges derives O(n²) quads in ~O(n³) time, and `getRulesFromDataset` reads rules straight from the graph — so a small semi-trusted document can exhaust CPU/memory, and a single rule with very many body triples overflows the `_evaluatePremise` recursion with an uncatchable RangeError.

Adds two opt-in constructor options (default unbounded, behaviour unchanged unless set):

- `maxDerivations` — caps genuinely-new derived quads; `reason()` throws a controlled error past the budget (measured: the 400-edge closure runs ~5 s / 80200 quads unbounded, fails closed in milliseconds at 1000).
- `maxPremiseDepth` — rejects up front any rule with more body triples than the premise recursion can take.

Consistency on throw — fixed from an earlier draft of this change, where a review bot caught an ordering bug: the budget check now runs after all three store indexes (`subjects`/`predicates`/`objects`) are updated for the derived quad, and the cached `_size` is invalidated in a `finally`, so catching the budget error leaves the store fully consistent (the reasoning result is merely incomplete). A regression test enumerates the store through the subject index and asserts every quad is also reachable through the predicate and object indexes — verified to fail if the check is placed between the index inserts.

Overhead when unset is noise: the unbounded 400-edge closure benchmark shows medians within 4% with overlapping ranges before/after (one increment + compare against Infinity per new derivation).

cc @jeswr

### DIFF
--- a/src/N3Reasoner.js
+++ b/src/N3Reasoner.js
@@ -16,14 +16,9 @@ export function getRulesFromDataset(dataset) {
 export default class N3Reasoner {
   constructor(store, options = {}) {
     this._store = store;
-    // Optional safety budgets for reasoning over rules and data that are not
-    // fully trusted. Naive forward chaining is super-linear: a transitive-closure
-    // rule over a chain of n edges derives O(n^2) quads in ~O(n^3) time, so a
-    // small input can exhaust CPU and memory. Both budgets default to unbounded
-    // for compatibility; `reason()` throws when a budget is exceeded.
+    // Optional safety budgets for reasoning over untrusted rules or data
     this._maxDerivations = options.maxDerivations === undefined ? Infinity : options.maxDerivations;
-    // `_evaluatePremise` recurses once per premise of a rule, so a single rule
-    // with very many body triples can overflow the stack. Bound the premise count.
+    // Caps a rule's premise count, as `_evaluatePremise` recurses once per premise
     this._maxPremiseDepth = options.maxPremiseDepth === undefined ? Infinity : options.maxPremiseDepth;
   }
 
--- a/src/N3Reasoner.js
+++ b/src/N3Reasoner.js
@@ -14,8 +14,17 @@ export function getRulesFromDataset(dataset) {
 }
 
 export default class N3Reasoner {
-  constructor(store) {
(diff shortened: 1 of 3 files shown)


src/N3Reasoner.js
```diff
@@ -14,15 +14,29 @@ export function getRulesFromDataset(dataset) {
 }
 
 export default class N3Reasoner {
-  constructor(store) {
+  constructor(store, options = {}) {
```
jeswr, line 17: Please make sure these params are documented in a readme.

src/N3Reasoner.js
```diff
@@ -14,15 +14,29 @@ export function getRulesFromDataset(dataset) {
...
     this._store = store;
+    // Optional safety budgets for reasoning over rules and data that are not
+    // fully trusted. Naive forward chaining is super-linear: a transitive-closure
+    // rule over a chain of n edges derives O(n^2) quads in ~O(n^3) time, so a
+    // small input can exhaust CPU and memory. Both budgets default to unbounded
+    // for compatibility; `reason()` throws when a budget is exceeded.
```
jeswr: comment too detailed

src/N3Reasoner.js
```diff
@@ -14,15 +14,29 @@ export function getRulesFromDataset(dataset) {
...
+    this._maxDerivations = options.maxDerivations === undefined ? Infinity : options.maxDerivations;
+    // `_evaluatePremise` recurses once per premise of a rule, so a single rule
+    // with very many body triples can overflow the stack. Bound the premise count.
```
jeswr: comment too long

jeswr:
Addressed: maxDerivations and maxPremiseDepth are now documented in the README (## Reasoning → "Limiting reasoning cost") with a worked example, and the two flagged constructor comments are trimmed to one line each — the rationale now lives in the README and the description rather than the code.

jeswr (commented):
LGTM

PR #658, opened 2026-07-04
fix(lexer): reject \U escapes beyond U+10FFFF

jeswr:
#593 rejects `\u`/`\U` escapes in the surrogate range, but `\U` escapes above U+10FFFF are still accepted and silently mangled: on main, `"\U00110000"` parses to a length-2 string of unpaired code units `0xDC00 0xDC00` (the surrogate-pair arithmetic applied beyond its domain). Code points above U+10FFFF are not Unicode scalar values (Turtle UCHAR / RFC 3987), so there is no character the escape could denote.

Reject `charCode > 0x10FFFF` in `_unescape`'s 8-digit branch, next to #593's surrogate check; the error is the same shape as the surrogate rejection (`Unexpected ""\U00110000"" on line 1.`). Two existing tests asserted the old over-accepting behaviour (expecting the mangled code units); they now use a valid astral escape (`\U0001F0A1`) instead. Added negative tests for the literal and IRI positions plus a boundary test for `\U0010FFFF`.

Note: this is the residual half of the `bad-numeric-escape` conformance skip (the surrogate half was #593); removing that skip can follow separately once this lands.

GIGO note: same class as #593 — accepting the escape can only ever produce corrupt output from N3's own unescaping.

cc @jeswr

### DIFF
--- a/src/N3Lexer.js
+++ b/src/N3Lexer.js
@@ -15,8 +15,9 @@ const escapeReplacements = {
 };
 const illegalIriChars = /[\x00-\x20<>\\"\{\}\|\^\`]/;
 
-function isSurrogateCodePoint(charCode) {
-  return charCode >= 0xD800 && charCode <= 0xDFFF;
+// A valid code point is a Unicode scalar value: at most U+10FFFF and not a surrogate
+function isValidCodePoint(charCode) {
+  return charCode <= 0x10FFFF && (charCode < 0xD800 || charCode > 0xDFFF);
 }
 
 const lineModeRegExps = {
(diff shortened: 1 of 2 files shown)


src/N3Lexer.js
```diff
@@ -434,7 +434,8 @@ export default class N3Lexer {
       // 8-digit unicode character
       if (typeof unicode8 === 'string') {
         let charCode = Number.parseInt(unicode8, 16);
-        if (isSurrogateCodePoint(charCode)) {
+        // Reject code points beyond U+10FFFF, which are not Unicode scalar values
+        if (charCode > 0x10FFFF || isSurrogateCodePoint(charCode)) {
```
jeswr: Should this check not be applied everywhere that `isSurrogateCodePoint`? Assuming that is the case why not make this into an `isValidCodePoint` function.

jeswr:
Addressed: the over-range and surrogate checks are unified into a single isValidCodePoint helper (a Unicode scalar value — at most U+10FFFF and not a surrogate), applied at both the \u and \U sites in _unescape. Behaviour unchanged; existing tests cover both rejection paths; the W3C spec suite is green.

jeswr (commented):
LGTM

PR #661, opened 2026-07-05
fix: escape IRIREF-forbidden characters and full control range in Writer (#659)

### LINKED ISSUE #659: Writer does not escape IRIs (RDF-injection) and emits raw C1/DEL control chars in literals
## Summary

`Writer` (Turtle/N-Triples/N-Quads/TriG) does two things that surprised us when serialising RDF whose term values originate from untrusted input:

1. **It does not escape/validate IRIs.** Whatever string a `NamedNode` carries is emitted verbatim between `<…>`. A value containing a Turtle-`IRIREF`-forbidden character (`>`, `<`, `"`, `{`, `}`, `|`, `^`, backtick, backslash, space, or a C0 control) breaks out of the angle brackets. With a crafted value this lets a **single intended quad serialise into two (or more) quads**, one of them entirely attacker-controlled — an RDF-injection primitive when any term value comes from parsed foreign RDF, a JSON/HTTP field, or a user-supplied URL.

2. **It emits several raw C1 / high C0 control characters into literal values.** `\t`, `\r`, `\b`, `\n` and `U+0000–U+0008` are correctly escaped, but `U+001A`, `U+001B` (ESC), `U+001F`, `U+007F` (DEL) and the C1 block (`U+0080`, `U+009B`, `U+009F`, …) pass through **raw** into the serialised literal. That is a terminal-escape / control-sequence smuggling vector in whatever consumes the output.

Node 25, `n3@2.1.0` (also observed on 1.x).

## Repro 1 — IRI injection (one quad in, two quads out)

```js
import { Writer, Parser, DataFactory } from "n3";
const { namedNode, quad } = DataFactory;

// A single attacker-controlled object IRI, crafted to close its own triple
// and forge a second, entirely attacker-chosen triple.
const hostile =
  "http://evil.example/o> . <http://victim.example/alice> " +
  "<http://xmlns.com/foaf/0.1/knows> <http://attacker.example/mallory";

const w = new Writer({ format: "text/turtle" });
w.addQuad(quad(
  namedNode("http://example.org/s"),
  namedNode("http://example.org/p"),
  namedNode(hostile),
));
const out = await new Promise((res) => w.end((_e, r) => res(r)));
console.log(out);
console.log("re-parsed quads:", new Parser().parse(out).length);
```

Output:

```
<http://example.org/s> <http://example.org/p> <http://evil.example/o> . <http://victim.example/alice> <http://xmlns.com/foaf/0.1/knows> <http://attacker.example/mallory>.
```

…which re-parses cleanly as **2 quads**, the second being the fully-forged
`<victim.example/alice> foaf:knows <attacker.example/mallory>`. The intended
graph had one quad.

## Repro 2 — raw control characters in a literal

```js
import { Writer, DataFactory } from "n3";
const { namedNode, literal, quad } = DataFactory;

const codes = [0x00,0x07,0x08,0x09,0x0a,0x0d,0x1a,0x1b,0x1f,0x7f,0x80,0x9b,0x9f];
const payload = codes.map((c) => String.fromCharCode(c)).join("|");

const w = new Writer({ format: "text/turtle" });
w.addQuad(quad(namedNode("http://example.org/s"), namedNode("http://example.org/p"), literal(payload)));
w.end((_e, out) => {
  for (const c of codes) {
    console.log(`U+${c.toString(16).toUpperCase().padStart(4,"0")} raw-in-output=${out.includes(String.fromCharCode(c))}`);
  }
});
```

Result — these survive **raw** in the serialised literal:

```
U+001A raw-in-output=true
U+001B raw-in-output=true   (ESC)
U+001F raw-in-output=true
U+007F raw-in-output=true   (DEL)
U+0080 raw-in-output=true
U+009B raw-in-output=true   (CSI)
U+009F raw-in-output=true
```

(`U+0000–U+0008`, `\t`, `\r`, `\n` are correctly escaped.)

## Expectation

- IRIs written by `Writer` should be validated and/or escaped so a term value can never break out of `<…>` (reject, or percent-encode the `IRIREF`-forbidden set `[#x00-#x20<>"{}|^` `` ` `` `\]`). At minimum, the writer should not emit output that misrepresents the input graph.
- Literal serialisation should escape the full C0 **and** C1 control ranges (not just the low C0 subset), so no raw ESC/DEL/CSI/C1 byte reaches the output.

## Workaround

Until then, callers serialising untrusted RDF must filter every IRI through an `IRIREF`-safe escaper and strip C1/DEL/high-C0 controls from literals before handing terms to `Writer`. We consolidated exactly this into `safeHttpIri` / `escapeIri` in [`@jeswr/rdf-serialize`](https://github.com/jeswr/rdf-serialize).

Happy to send a PR if the maintainers agree on the desired behaviour (validate-and-throw vs. escape) for each of the two.

---

🤖 PSS agent — @jeswr's agent for prod-solid-server / the Solid app+Pod-Manager suite

jeswr:
> 🤖 Filed by @jeswr's **PSS agent** (Claude); @jeswr runs multiple agents on this account — the **SPARQ** agent is a separate party in these discussions.

**Draft** — addresses #659. `N3Writer` emitted `NamedNode` values verbatim between angle brackets, so a value containing an `IRIREF`-forbidden character (`>`, `<`, `"`, `{`, `}`, `|`, `^`, backtick, `\`, space, or a C0 control) closed its own `<…>` and injected forged triples (repro 1 in the issue). Separately, literal escaping only covered U+0000–U+0019, so U+001A–U+001F, DEL (U+007F) and the C1 block (U+0080–U+009F) passed through raw (repro 2).

The fix (`src/N3Writer.js`):

1. **IRIs** — `_encodeIriOrBlank` (and `@prefix` lines) escape the `IRIREF`-forbidden set (U+0000–U+0020 and ``< > " { } | ^ ` \``) as UCHAR (`\uXXXX`/`\UXXXXXXXX`), via a dedicated replacer because `IRIREF` permits only UCHAR escapes, not ECHAR.
2. **Literals** — `_encodeLiteral` widens the escaped ranges to the full C0 plus DEL and C1 (U+0000–U+001F, U+007F–U+009F).

Why escape (UCHAR) rather than percent-encode or throw: percent-encoding changes IRI identity (`>` → `%3E` doesn't reverse), whereas UCHAR escaping round-trips byte-identically; a valid IRI contains no forbidden characters, so well-formed data is byte-for-byte unchanged; and re-reading escaped malformed input with N3's own strict parser rejects the invalid IRI instead of materialising an injected quad. If validate-and-throw is preferred instead, that's a small swap — the main reason this is a draft. (DEL/C1 are grammar-permitted raw inside an `IRIREF`, so they're only escaped in literals; escaping them in IRIs too is a one-line range addition if wanted.)

Tests: a `Writer term-serialisation safety (#659)` block covers the issue's repros (re-parsing never yields the forged quad, an injected newline can't start a statement), every forbidden character emitted as UCHAR, valid IRIs (`%20`, query, fragment, punycode host, `urn:`, `did:`) byte-for-byte unchanged, astral `\U…` behaviour preserved, the full C0+DEL+C1 set in literals, and a hostile `@prefix`. `npm test` green (6512 tests, 100% coverage on `N3Writer.js`), lint clean.

Closes #659

### DIFF
--- a/src/N3Writer.js
+++ b/src/N3Writer.js
@@ -9,13 +9,29 @@ const DEFAULTGRAPH = N3DataFactory.defaultGraph();
 
 const { rdf, xsd } = namespaces;
 
-// Characters in literals that require escaping
-const escape    = /["\\\t\n\r\b\f\u0000-\u0019\ud800-\udbff]/,
-    escapeAll = /["\\\t\n\r\b\f\u0000-\u0019]|[\ud800-\udbff][\udc00-\udfff]/g,
+// Characters in literals that require escaping: the full C0 control range
+// U+0000-U+001F (which includes \t \n \r \b \f) plus DEL and the C1 control
+// range U+007F-U+009F, as well as the quote and backslash. Escaping the full
+// control ranges (not only the low U+0000-U+0019 subset) keeps raw ESC / DEL /
+// CSI / C1 bytes -- a terminal-escape smuggling vector -- out of literals.
+const escape    = /["\\\t\n\r\b\f\u0000-\u001f\u007f-\u009f\ud800-\udbff]/,
+    escapeAll = /["\\\t\n\r\b\f\u0000-\u001f\u007f-\u009f]|[\ud800-\udbff][\udc00-\udfff]/g,
     escapedCharacters = {
       '\\': '\\\\', '"': '\\"', '\t': '\\t',
       '\n': '\\n', '\r': '\\r', '\b': '\\b', '\f': '\\f',
     };
+
+// Characters in IRIs that require escaping. The Turtle / N-Triples IRIREF
+// grammar permits inside <...> only the set  [^#x00-#x20<>"{}|^`\]
+// together with UCHAR numeric escapes; there is NO \n / \t / \" -style
+// ECHAR in an IRIREF. So the control-or-space range U+0000-U+0020 and each of
+// the symbols  < > " { } | ^ ` (backtick) \  must be written as a
+// UCHAR escape (\uXXXX / \UXXXXXXXX). Without this a NamedNode value
+// carrying a > (or a newline, space, ...) closes its own <...> early and
+// injects attacker-controlled triples (RDF-injection). Astral characters and
+// lone surrogates are escaped too, so no raw such code unit reaches the output.
+const escapeIri    = /[\u0000-\u0020<>"{}|^`\\]|[\ud800-\udbff][\udc00-\udfff]|[\ud800-\udfff]/,
+    escapeIriAll = /[\u0000-\u0020<>"{}|^`\\]|[\ud800-\udbff][\udc00-\udfff]|[\ud800-\udfff]/g;
 
 // ## Placeholder class to represent already pretty-printed terms
 class SerializedTerm extends Term {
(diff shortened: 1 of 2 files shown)


jeswr (commented):
We need to discuss the following RDF/JS assumptions before this PR is merged:
 - Do we assume named node IRI's should be pre-escaped

We need to discuss the following N3.js assumption:
 - Do we assume prefixes to be sanitised

The following N3.js assumption is known: We expect named nodes to be valid RDF/JS terms.

PR #662, opened 2026-07-05
perf(store): remove per-call allocations from index read hot paths

jeswr:
Removes per-call allocations from the `N3Store` index read hot paths. A single commit on `main`, touching only `src/N3Store.js`; behaviour-preserving (full suite passes unchanged at 100% coverage).

**What it does**

- `N3EntityIndex._termFromNumericId` + `_termCache`: memoizes the materialized term per numeric entity id. Entities are never removed from `_ids`/`_entities`, so cached terms cannot go stale. Repeated reads reuse one term instance instead of re-materializing per result; memory cost is one term per distinct entity actually read. This is safe under the RDF/JS contract — terms are immutable (the data-model spec requires it, with equivalence by `equals()` not identity), and the read path already shares instances today (the `DefaultGraph` singleton; the graph/outer terms within a single `_findInIndex` call).
- `_findInIndex` / `_countInIndex`: stops allocating a throwaway single-key object (`{ [key]: ... }` + `for`-`in`) at every bound index level on every call, iterating the bound level directly; `_countInIndex` is split into monomorphic per-level helpers.

**Why**: CPU profiles of read-heavy downstream workloads (Comunica SPARQL over an in-memory store, per-binding property lookups on a 38k-quad graph, CSS parse/serialize round-trips) showed `_findInIndex` as the dominant read self-time site, much of it term re-materialization and the garbage it produces.

**Measured** — `node perf/N3Store-perf.js 128`, this branch vs its merge base, medians of 5 interleaved runs per side, Node v25.1.0, Apple M1:

| scenario | main | this PR | speedup |
|---|---|---|---|
| fully-bound getQuads (0 variables) | 2.80 s | 1.19 s | 2.35x |
| 1-variable finds | 677 ms | 552 ms | 1.23x |
| 2-variable finds | 639 ms | 553 ms | 1.16x |
| all-bound quad finds | 463 ms | 399 ms | 1.16x |
| single-match by one term (×1M) | 1.09 s | 0.84 s | ~1.3x |
| single-match by two/three terms (×1M) | 1.21 s | 0.82 s | ~1.5x |
| full scan (all quads) | 520 ms | 484 ms | 1.07x |
| adds / RSS | — | — | parity |

Minor-GC count over one full perf-script run drops from 658 to 442 scavenges (`--trace-gc`); major GC unchanged.

**Relationship to #635**: an earlier version was stacked on #635, so its diff also showed #635's commits (the cross-index set-ops, the component-id triple-term representation, `perf/N3StoreSetOps-perf.js`), and its numbers were measured against that stack. It has been de-stacked and re-measured: the diff is now exactly the read-path change above, all numbers are against plain `main`. The two changes are independent but compound — under #635's component-id triple-term representation term materialization becomes a recursive rebuild, which the memoization here avoids on repeated reads. The design questions raised on the old diff (WeakMap for the shared index, RDF 1.2 native-indexing overlap) concern #635's content and are best discussed there.

### DIFF
--- a/src/N3Store.js
+++ b/src/N3Store.js
@@ -90,6 +90,10 @@ export class N3EntityIndex {
      // inverse of `_ids`
     this._entities = Object.create(null);
     this._entities[1] = '';
+    // `_termCache` memoizes the materialized term per numeric entity id.
+    // Entities are never removed from `_ids`/`_entities`, so entries never go stale,
+    // and RDF/JS terms are immutable, so instances can be shared across reads.
+    this._termCache = [];
     // `_blankNodeIndex` is the index of the last automatically named blank node
     this._blankNodeIndex = 0;
     this._factory = options.factory || N3DataFactory;
@@ -108,6 +112,12 @@ export class N3EntityIndex {
       return q;
     }
     return termFromId(id, this._factory);
+  }
+
+  // ### `_termFromNumericId` materializes the term with the given numeric id,
+  // caching the result so repeated reads reuse a single term instance
+  _termFromNumericId(id) {
+    return this._termCache[id] || (this._termCache[id] = this._termFromId(this._entities[id]));
   }
 
   _termToNumericId(term) {
@@ -170,6 +180,7 @@ export default class N3Store {
     this._entityIndex = options.entityIndex || new N3EntityIndex({ factory: this._factory });
     this._entities = this._entityIndex._entities;
     this._termFromId = this._entityIndex._termFromId.bind(this._entityIndex);
+    this._termFromNumericId = this._entityIndex._termFromNumericId.bind(this._entityIndex);
     this._termToNumericId = this._entityIndex._termToNumericId.bind(this._entityIndex);
     this._termToNewNumericId = this._entityIndex._termToNewNumericId.bind(this._entityIndex);
 
@@ -234,26 +245,27 @@ export default class N3Store {
   // (for instance: _subject_, _predicate_, and _object_).
   // Finally, `graphId` will be the graph of the created quads.
   *_findInIndex(index0, key0, key1, key2, name0, name1, name2, graphId) {
-    let tmp, index1, index2;
-    const entityKeys = this._entities;
-    const graph = this._termFromId(entityKeys[graphId]);
+    let index1, index2;
+    const graph = this._termFromNumericId(graphId);
     const parts = { subject: null, predicate: null, object: null };
 
     // If a key is specified, use only that part of index 0.
-    if (key0) (tmp = index0, index0 = {})[key0] = tmp[key0];
-    for (const value0 in index0) {
+    const values0 = key0 ? [key0] : Object.keys(index0);
+    for (let i0 = 0; i0 < values0.length; i0++) {
+      const value0 = values0[i0];
       if (index1 = index0[value0]) {
-        parts[name0] = this._termFromId(entityKeys[value0]);
+        parts[name0] = this._termFromNumericId(value0);
         // If a key is specified, use only that part of index 1.
-        if (key1) (tmp = index1, index1 = {})[key1] = tmp[key1];
-        for (const value1 in index1) {
+        const values1 = key1 ? [key1] : Object.keys(index1);
+        for (let i1 = 0; i1 < values1.length; i1++) {
+          const value1 = values1[i1];
           if (index2 = index1[value1]) {
-            parts[name1] = this._termFromId(entityKeys[value1]);
+            parts[name1] = this._termFromNumericId(value1);
             // If a key is specified, use only that part of index 2, if it exists.
             const values = key2 ? (key2 in index2 ? [key2] : []) : Object.keys(index2);
             // Create quads for all items found in index 2.
             for (let l = 0; l < values.length; l++) {
-              parts[name2] = this._termFromId(entityKeys[values[l]]);
+              parts[name2] = this._termFromNumericId(values[l]);
               yield this._factory.quad(parts.subject, parts.predicate, parts.object, graph);
             }
           }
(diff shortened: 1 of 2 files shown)


perf/N3StoreSetOps-perf.js
jeswr, line 1: Is there not already a performance test like this. Can we not update it rather than creating new ones?

src/N3Store.js
```diff
@@ -17,6 +17,20 @@ function merge(target, source, depth = 4) {
...
+ */
+function addToIndex(index0, key0, key1, key2) {
+  // Create layers as necessary
+  const index1 = index0[key0] || (index0[key0] = {});
+  const index2 = index1[key1] || (index1[key1] = {});
+  // Setting the key to _any_ value signals the presence of the quad
+  const existed = key2 in index2;
+  if (!existed)
+    index2[key2] = null;
+  return !existed;
```
```suggestion
  return false;
```

src/N3Store.js
```diff
@@ -79,6 +93,174 @@ function difference(s1, s2, depth = 4) {
...
+ * Only entities present in both indices appear in the returned map.
+ *
+ * Composite (triple term) ids are remapped via their component-id arrays.
+ * A composite is interned after its components, so its components are always
+ * already remapped when it is reached: ids are integer-index keys below
+ * 2^32 - 1, which `Object.keys` enumerates in ascending numeric order.
+ */
+function remapEntityIds(source, target) {
+  const entities = source._entities, targetIds = target._ids;
+  const remap = Object.create(null);
```
```suggestion
  const remap = {};
```

Why `Object.create(null)` here and actual objects in `addToIndex`?

src/N3Store.js
```diff
@@ -79,6 +93,174 @@ function difference(s1, s2, depth = 4) {
...
+/**
+ * Maps the numeric ids of the `source` entity index to those of `target`.
+ * Only entities present in both indices appear in the returned map.
+ *
+ * Composite (triple term) ids are remapped via their component-id arrays.
+ * A composite is interned after its components, so its components are always
+ * already remapped when it is reached: ids are integer-index keys below
+ * 2^32 - 1, which `Object.keys` enumerates in ascending numeric order.
+ */
+function remapEntityIds(source, target) {
```
jeswr: Not sure this is worth it. 
1. We should first investigate whether something like `WeakMap` can instead be used to enable all stores to share memory, without long term having them hold on to all references, and without a performance loss.
2. If an application really cares that much about performance they can follow the docs to make two stores share an index.

src/N3Store.js
```diff
@@ -79,6 +93,174 @@ function difference(s1, s2, depth = 4) {
...
+ * where quads whose graph/subject/predicate/object has no counterpart in
+ * `g2`'s index are exactly the ones to keep, so `undefined` is carried down
+ * through all four levels instead of skipping.
+ *
+ * NOTE: unlike the depth-recursive `intersect`/`difference`, the descent is
+ * hand-inlined (like the `size` getter, `_findInIndex` and `_countInIndex`)
+ * because each level must consult `remap`; sharing a generic walker with
+ * those hot monomorphic per-quad loops would slow them down.
+ */
+function crossGraphsOp(g1, g2, remap, keepIfPresent) {
```
jeswr: See https://github.com/rdfjs/N3.js/pull/662/changes#r3546083456

src/N3Store.js
```diff
@@ -79,6 +93,174 @@ function difference(s1, s2, depth = 4) {
...
+/**
+ * Cross-entity-index intersection that walks the smaller `_graphs` index `g1`
+ * while probing the larger `g2`: like `crossGraphsOp` with
+ * `keepIfPresent === true`, except that unmapped ids prune the walk, and kept
+ * quads are emitted under their remapped ids so the returned `graphs` index is
+ * keyed by `g2`'s ids. This is only valid for intersection: `remap` is
+ * injective, so the remapped quads are still unique per graph, whereas
+ * difference must emit the unmappable quads of the walked index.
+ */
+function crossGraphsIntersect(g1, g2, remap) {
```
jeswr: See https://github.com/rdfjs/N3.js/pull/662/changes#r3546083456

src/N3Store.js
```diff
@@ -79,6 +93,174 @@ function difference(s1, s2, depth = 4) {
...
+  return { graphs: target, size };
+}
+
+/**
+ * Shared shell for the cross-entity-index branch of `intersection`
+ * (`keepIfPresent === true`) and `difference` (`keepIfPresent === false`):
+ * builds a result store keyed by `self`'s entity index, remapping the ids of
+ * the walked operand into the id space of the probed one.
+ */
+function crossIndexOp(self, other, keepIfPresent) {
```
jeswr: See https://github.com/rdfjs/N3.js/pull/662/changes#r3546083456

src/N3Store.js
```diff
@@ -90,47 +272,100 @@ export class N3EntityIndex {
      // inverse of `_ids`
     this._entities = Object.create(null);
     this._entities[1] = '';
+    // `_quadIds` maps triple terms to numeric ids through their components:
+    // `sId -> pId -> oId -> termId`. The object level holds the term id
+    // directly for a default-graph term (the only kind the parser produces)
+    // and spills to a graph sub-map `graphId -> termId` (with 1 for the
+    // default graph) once a graph-component term is interned for that triple
+    this._quadIds = Object.create(null);
```
jeswr: This overlaps with / is made redundant by the work to have RDF 1.2 native indexing.

jeswr:
Reworked. The concrete asks are applied and the contentious pieces resolve by de-stacking:

- The lines flagged at :31, :105, :107, :146/:197/:244 and :280, and the new perf/N3StoreSetOps-perf.js, were all #635 content that appeared here only because this PR was stacked on #635. It is now de-stacked to a single commit on main, so the diff is exactly the read-path allocation change its title names — that code is no longer part of it. The WeakMap and RDF-1.2-native-indexing design questions concern #635 and are best discussed there.
- The perf numbers were re-measured against plain main (the earlier figures were carried over from the #635 stack): fully-bound getQuads 2.80s to 1.19s, single/two-term retrievals ~1.3-1.5x, adds and RSS at parity, minor-GC scavenges down ~33%. Full table in the updated description.
- The term memoization shares one immutable term instance per entity across reads — within the RDF/JS contract (terms are immutable, equivalence by equals() not identity), and the read path already shares instances today (the DefaultGraph singleton, the outer terms within one _findInIndex). It cannot poison the index, which stores numeric ids rather than term references; a test pins that repeated reads keep returning terms equal to freshly created ones.

jeswr:
Lets pause work on this for now and turn it back into an issue. We should land RDF 1.2 indexing first; investigate the option of 'lazy' materialisation of quads; and use a WeakMap for this kind of caching if we do pursue it to avoid memory overload issues.

PR #664, opened 2026-07-06
fix(writer): represent blank nodes through their term type

jeswr:
`N3Writer._encodeIriOrBlank` decided whether a non-`NamedNode` term was already serialized by probing for an `id` property:

```js
return 'id' in entity ? entity.id : `_:${entity.value}`;
```

A `BlankNode` from another RDF/JS implementation may carry an `id` property with unrelated content, which was then written into the document verbatim — e.g. `{ termType: 'BlankNode', value: 'b1', id: 'anything at all' }` serialized as `anything at all`, an output-integrity hazard. `_encodeIriOrBlank` now dispatches blank nodes on `termType` (`_:` + `value`), like the rest of the writer. N3.js's own blank-node terms are unaffected: their `id` and the `termType`-derived form are identical, as covered by the existing suite (100% coverage).

The `Variable` case is a separate change and now lives in #667.

> **Review timing:** prepared with Claude; I (@jeswr) will personally review before it progresses.

### DIFF
--- a/src/N3Writer.js
+++ b/src/N3Writer.js
@@ -151,7 +151,7 @@ export default class N3Writer {
       // If it is a list head, pretty-print it
       if (this._lists && (entity.value in this._lists))
         entity = this.list(this._lists[entity.value]);
-      return 'id' in entity ? entity.id : `_:${entity.value}`;
+      return entity.termType === 'BlankNode' ? `_:${entity.value}` : entity.id;
     }
     let iri = entity.value;
     // Use relative IRIs if requested and possible
--- a/src/N3Writer.js
+++ b/src/N3Writer.js
@@ -151,7 +151,11 @@ export default class N3Writer {
       // If it is a list head, pretty-print it
       if (this._lists && (entity.value in this._lists))
         entity = this.list(this._lists[entity.value]);
-      return entity.termType === 'BlankNode' ? `_:${entity.value}` : entity.id;
+      if (entity.termType === 'BlankNode')
+        return `_:${entity.value}`;
+      if (entity.termType === 'Quad')
+        return this._encodeQuad(entity);
+      return entity.id;
     }
     let iri = entity.value;
     // Use relative IRIs if requested and possible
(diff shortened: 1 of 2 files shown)


src/N3Writer.js
```diff
@@ -151,7 +151,11 @@ export default class N3Writer {
       // If it is a list head, pretty-print it
       if (this._lists && (entity.value in this._lists))
         entity = this.list(this._lists[entity.value]);
-      return 'id' in entity ? entity.id : `_:${entity.value}`;
+      // Represent the term through its term type,
+      // since terms from other libraries could use `id` for other purposes
```
jeswr: This comment belongs in the PR description not a coment, it refers to a code idiosyncrasy being removed.

src/N3Writer.js
```diff
@@ -151,7 +151,11 @@ export default class N3Writer {
...
+      return entity.termType === 'BlankNode' ? `_:${entity.value}` :
+             entity.termType === 'Variable' ? `?${entity.value}` :
```
jeswr: The variable case is introduced in this PR. This should be a separate PR as this is a different change.

jeswr:
Addressed: reduced to the blank-node change only. The removed id-probe rationale now lives in the description (no code comment narrates it), and the variable case is split into #667.

src/N3Writer.js
```diff
@@ -151,7 +151,7 @@ export default class N3Writer {
       // If it is a list head, pretty-print it
       if (this._lists && (entity.value in this._lists))
         entity = this.list(this._lists[entity.value]);
-      return 'id' in entity ? entity.id : `_:${entity.value}`;
+      return entity.termType === 'BlankNode' ? `_:${entity.value}` : entity.id;
```
jeswr: There is still a bug here if `entity` is a quad that is not minted by the N3.js library.

I guest we need to add a check to see if `entity` is constructed by this library - and if not apply full serialisation logic.

jeswr: We should also use something like `it.each` to run the tests suit with both an N3 data factory; and with a datafactory from a different library. That should be done as a follow up PR.

PR #640, opened 2026-07-02
chore(tests): run the W3C Notation3 spec suites in CI

jeswr:
Runs the W3C Notation3 parser and extended suites (`manifest-parser.ttl` and `manifest-extended.ttl`) against N3.js as `spec-n3-parser` and `spec-n3-extended` jobs in the CI spec matrix. No project-maintained skip list is applied: this PR remains draft and the jobs deliberately expose unsupported cases. Closes #331.

The suites moved to the Notation3 CG, so their canonical manifests are under `https://w3c-cg.github.io/N3/tests/N3Tests/` (the old `w3c.github.io/N3/tests/…` URLs now return 404). `manifest-extended.ttl` contains parser syntax tests rather than reasoning or evaluation tests.

**Why a local checkout + `-m` instead of `-c`**: rdf-test-suite constructs all test cases concurrently, producing roughly 1,000 simultaneous GitHub Pages fetches and reliable `UND_ERR_CONNECT_TIMEOUT` failures. PR #612 makes warm `.rdf-test-suite-cache` runs reliable but cannot safely populate a cold N3 cache. `spec-n3-fetch` therefore creates a sparse, blob-filtered checkout of `w3c-cg/N3` and the suites use rdf-test-suite mapping from URLs to files, preserving their canonical test IRIs without the request burst.

The fixture checkout is pinned to the N3 revision in `spec/cache-key.txt`. The weekly refresh workflow checks `w3c-cg/N3` alongside the RDF 1.1/1.2 suites and proposes revision changes for review, so test updates do not arrive silently.

**Current results at the pinned revision**: parser 173/222 pass (16 suite-rejected tests skipped); extended 974/978 pass (107 suite-rejected tests skipped). Two parser-manifest `TestN3Eval` entries are not constructed because the released rdf-test-suite package cannot load their expected-result files.

**Validated dependency stack**: parser 223/223 pass (16 skipped); extended 978/978 pass (107 skipped). The regular N3.js suite also passes: 20/20 suites and 6,861/6,861 tests with 100% coverage, plus lint.

## Merge dependencies

N3.js parser support:

- [ ] #693 — opt-in implicit empty prefix, enabled by the conformance adapter
- [ ] #694 — `has`, `is ... of`, and `<-` inverse predicates
- [ ] #695 — IRI property lists
- [ ] #696 — expression predicates
- [ ] #697 — directives inside formulas

Upstream test infrastructure and fixtures:

[...]

PR #693, opened 2026-08-24
feat(parser): add opt-in implicit N3 empty prefix

### LINKED ISSUE #687: N3 parser does not provide the historical implicit empty prefix
In `text/n3` mode, prefixed names using the empty prefix fail unless `@prefix : ...` is declared explicitly:

```js
new Parser({ format: 'text/n3', baseIRI: 'https://example.org/doc.n3' })
  .parse(':s :p :o .');
// Undefined prefix ":" on line 1
```

The W3C N3 parser and extended manifests currently rely on the historical Team Submission convention that the empty prefix is implicitly bound to `<#>`; this accounts for 32 failures exposed by #640. A 2023 prototype exists as `4f5ffbf` on `feat/full-n3-support`, but it was never merged and should be reimplemented using the current IRI resolver (including bases that already contain fragments).

The normative status is being clarified upstream in w3c-cg/N3#235. A draft implementation here would make the compatibility choice explicit and give us something concrete to review.

Part of #640.

jeswr:
Adds an opt-in `implicitEmptyPrefix` parser option for the historical N3 convention where an undeclared `:` prefix means `<#>` relative to the document IRI. The conformance adapter enables the option; normal parser behavior is unchanged in this major version. An explicit `@prefix : ...` still overrides it, and N3 input without a document IRI remains strict.

The binding uses the existing IRI resolver. Resolving `#` replaces a fragment on the supplied document IRI in accordance with RFC 3986 section 5.2: https://www.rfc-editor.org/rfc/rfc3986#section-5.2

This makes 26 additional parser-manifest cases and two additional extended-manifest cases pass in #640. The normative test/spec alignment is being clarified in w3c-cg/N3#235, and changing the option default in the next major release is tracked in #699.

Tests: full Jest suite (6,817 tests, 100% coverage); ESLint; focused W3C manifests.

Closes #687

### DIFF
--- a/test/N3Parser-test.js
+++ b/test/N3Parser-test.js
@@ -2662,29 +2662,40 @@ describe('Parser', () => {
 
   describe('A Parser instance for the N3 format', () => {
     function parser() { return new Parser({ baseIRI: BASE_IRI, format: 'N3' }); }
-    function parserWithFragment() { return new Parser({ baseIRI: 'http://example.com/doc#old', format: 'N3' }); }
+    function implicitEmptyPrefixParser() {
+      return new Parser({ baseIRI: BASE_IRI, format: 'N3', implicitEmptyPrefix: true });
+    }
+    function parserWithFragment() {
+      return new Parser({ baseIRI: 'http://example.com/doc#old', format: 'N3', implicitEmptyPrefix: true });
+    }
     function parserIsImpliedBy() { return new Parser({ baseIRI: BASE_IRI, format: 'N3', isImpliedBy: true }); }
 
     it(
       'should bind the empty prefix to the document local namespace',
-      shouldParse(parser, ':a :b :c .',
+      shouldParse(implicitEmptyPrefixParser, ':a :b :c .',
                   ['http://example.org/#a', 'http://example.org/#b', 'http://example.org/#c']),
     );
 
     it(
       'should let an explicit empty prefix override the implicit binding',
-      shouldParse(parser, '@prefix : <http://example.com/>. :a :b :c .',
+      shouldParse(implicitEmptyPrefixParser, '@prefix : <http://example.com/>. :a :b :c .',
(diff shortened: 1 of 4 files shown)


test/N3Parser-test.js
```diff
@@ -2662,8 +2662,32 @@ describe('Parser', () => {
...
+      'should let an explicit empty prefix override the implicit binding',
+      shouldParse(parser, '@prefix : <http://example.com/>. :a :b :c .',
+                  ['http://example.com/a', 'http://example.com/b', 'http://example.com/c']),
+    );
+
+    it(
+      'should replace a document IRI fragment in the implicit binding',
+      shouldParse(parserWithFragment, ':a :b :c .',
+                  ['http://example.com/doc#a', 'http://example.com/doc#b', 'http://example.com/doc#c']),
+    );
```
jeswr, line 2690: Please provide reference indicating that this is teh correct behavior - i.e. the stripping of the `old` from the baseIRI

jeswr (commented):
This behavior should be opt-in until the next major version. Please make it configurable and then open an issue to change the default in the next major release.

test/N3Parser-test.js
jeswr: Addressed in dacd247. The fragment handling follows RFC 3986 section 5.2: resolving `#` replaces the base fragment (https://www.rfc-editor.org/rfc/rfc3986#section-5.2). I added that reference beside the regression test and to the PR description.

jeswr:
Addressed the review in dacd247:

- made implicit empty-prefix binding opt-in via `implicitEmptyPrefix`
- enabled it only in the conformance adapter
- documented and tested the unchanged default
- opened #699 to enable it by default in the next major release
- added the RFC 3986 section 5.2 reference for fragment replacement

Validation: ESLint and all 6,817 Jest tests pass with 100% coverage.

PR #695, opened 2026-08-24
feat(parser): support N3 IRI property lists

### LINKED ISSUE #689: N3 parser does not support IRI property lists
The current N3 grammar defines `iriPropertyList ::= IPLSTART iri predicateObjectList ']'`. The five positive cases in the W3C parser manifest all fail today:

```n3
:s :p [id :identified :name "value"] .
```

The construct uses the supplied IRI as the property-list subject instead of minting a blank node. Negative cases also require the parser to distinguish illegal semicolons and non-IRI identifiers.

An old prototype exists as `04c1e74` on `feat/full-n3-support`, but it lacks current error propagation, complete lexer boundaries, and parser coverage.

Part of #640.

jeswr:
Implements the current N3 `iriPropertyList` production used by the W3C parser manifest:

```n3
:s :p [id :identified :name "value"] .
```

The supplied IRI replaces the provisional blank node in the enclosing subject, predicate, or object context before properties are emitted. Nested IRI property lists therefore preserve both inner and outer references. The parser requires an actual IRI after `id`, requires a predicate-object list, and keeps the suite's illegal-semicolon, multiple-ID, and blank-node-ID cases negative.

The lexer recognizes `id` only in N3 mode, waits across stream chunks, and preserves `id:p` as a normal prefixed name.

This makes all five positive IRI-property-list cases in #640 pass.

Tests: full Jest suite (6,825 tests, 100% coverage); ESLint; W3C parser manifest improves from 173/222 to 178/222 on this branch alone.

Closes #689

### DIFF
--- a/src/N3Lexer.js
+++ b/src/N3Lexer.js
@@ -57,6 +57,7 @@ export default class N3Lexer {
     this._boolean = /^(?:true|false)(?=[.,;!\^\s#()\[\]\{\}"'<>])/;
     this._atKeyword = /^@[a-z]+(?=[\s#<:])/i;
     this._keyword = /^(?:PREFIX|BASE|VERSION|GRAPH)(?=[\s#<])/i;
+    this._n3Id = /^id(?=[\s#<])/;
     this._shortPredicates = /^a(?=[\s#()\[\]\{\}"'<>])/;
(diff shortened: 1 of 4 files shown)


jeswr:
LGTM - just awaiting copilot comment

PR #702, opened 2026-08-24
docs: Donating to contributors

RubenVerborgh:
While I was the original author, others have taken on the work since. Their efforts should be reflected correctly.


jeswr (approved):
Thankyou 😄 🚀

PR #703, opened 2026-08-24
fix(writer): sanitize IRI escaping, validate variable serialization, and fix TriG annotation scoping (#685, #684, #659)

### LINKED ISSUE #659: Writer does not escape IRIs (RDF-injection) and emits raw C1/DEL control chars in literals

### LINKED ISSUE #684: Remove index entry counters when JavaScript engines optimize empty-index checks
PR #647 adds Symbol-keyed entry counters to N3Store index nodes so removal can detect empty layers without enumerating their properties. This works around the JavaScript-engine behavior tracked in [Chromium issue 530857250](https://issues.chromium.org/issues/530857250).

Revisit the counters once supported JavaScript engines optimize this pattern. Before removing them, benchmark indexed removal and confirm that property enumeration (or an equivalent empty-object check) no longer regresses large index nodes. The cleanup should simplify the counter maintenance in add/remove, merge/intersection/difference, and matched-index construction while preserving the constant-time count optimizations where useful.

Follow-up to #647.

### LINKED ISSUE #685: Make sure tracked `size` is used to help optimise operations
In https://github.com/rdfjs/N3.js/pull/647/changes `SIZE` is introduced to track index sizes. This can be used to help work out e.g. where the outer loop should be on merge and intersect operations

Thundercloud12:
### Summary of Changes
Fixes #685, fixes #684, fixes #659.

1. **IRI & Control Character Escaping (Fixes #685)**:
   - Added regex sanitization for IRIs and extended control character ranges (`C0`/`C1`) in `N3Writer.js` to prevent RDF injection attacks.

2. **Variable Term Validation (Fixes #684)**:
   - Added validation in `N3Writer.js` to prevent `Variable` terms (`?var`) from being serialized into formats that do not support variables (e.g., Turtle, N-Triples).

3. **TriG Graph Scoping for Annotations (Fixes #659)**:
   - Updated `N3Parser.js` to ensure synthesized `rdf:reifies` quads inherit the active named graph context in TriG annotation syntax instead of defaulting to the default graph.

### Verification & Testing
- Added comprehensive unit test coverage in `test/N3Writer-test.js` and `test/N3Parser-test.js`.
- Verified 100% test coverage across statements, branches, functions, and lines (`npm test` — all 6,817 tests passing).
- Checked linting (`npm run lint`).

### DIFF
--- a/src/N3Writer.js
+++ b/src/N3Writer.js
@@ -10,8 +10,9 @@ const DEFAULTGRAPH = N3DataFactory.defaultGraph();
 const { rdf, xsd } = namespaces;
 
 // Characters in literals that require escaping
-const escape    = /["\\\t\n\r\b\f\u0000-\u0019\ud800-\udbff]/,
-    escapeAll = /["\\\t\n\r\b\f\u0000-\u0019]|[\ud800-\udbff][\udc00-\udfff]/g,
+const escape    = /["\\\t\n\r\b\f\u0000-\u001f\u007f-\u009f\ud800-\udbff]/,
+    escapeAll = /["\\\t\n\r\b\f\u0000-\u001f\u007f-\u009f]|[\ud800-\udbff][\udc00-\udfff]/g,
+    escapeIri = /[<>"{}|^`\\\s\u0000-\u001f\u007f-\u009f]|[\ud800-\udbff][\udc00-\udfff]/g,
     escapedCharacters = {
       '\\': '\\\\', '"': '\\"', '\t': '\\t',
       '\n': '\\n', '\r': '\\r', '\b': '\\b', '\f': '\\f',
@@ -37,6 +38,7 @@ export default class N3Writer {
     if (outputStream && typeof outputStream.write !== 'function')
       options = outputStream, outputStream = null;
     options = options || {};
+    this._format = options.format || '';
     this._lists = options.lists;
 
     // If no output stream given, send the output as string through the end callback
@@ -115,14 +117,23 @@ export default class N3Writer {
                     this._encodePredicate(this._predicate = predicate)} ${
                     this._encodeObject(object)}`, done);
     }
-    catch (error) { done && done(error); }
+    catch (error) {
+      if (done) done(error);
+      else throw error;
+    }
   }
 
   // ### `_writeQuadLine` writes the quad to the output stream as a single line
   _writeQuadLine(subject, predicate, object, graph, done) {
     // Write the quad without prefixes
     delete this._prefixMatch;
-    this._write(this.quadToString(subject, predicate, object, graph), done);
+    try {
+      this._write(this.quadToString(subject, predicate, object, graph), done);
+    }
+    catch (error) {
+      if (done) done(error);
+      else throw error;
+    }
   }
 
   // ### `quadToString` serializes a quad as a string
@@ -154,8 +165,12 @@ export default class N3Writer {
       // If it is a list head, pretty-print it
       if (this._lists && (entity.value in this._lists))
         entity = this.list(this._lists[entity.value]);
-      return entity.termType === 'Variable' ? `?${entity.value}` :
-             'id' in entity ? entity.id : `_:${entity.value}`;
+      if (entity.termType === 'Variable') {
+        if (this._format && !(/n3|notation3/i).test(this._format))
+          throw new Error(`Cannot write Variable term in ${this._format} format.`);
+        return `?${entity.value}`;
+      }
+      return 'id' in entity ? entity.id : `_:${entity.value}`;
     }
     let iri = entity.value;
     // Use relative IRIs if requested and possible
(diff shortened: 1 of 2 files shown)


Thundercloud12:
@jeswr Hey so just started out in open source contribution so this issues that i thought would be good for me to work on, can you review my pr?

Thundercloud12:
@westurner @antoniogarrote can you have a look at this pr?

jeswr:
Hi @Thundercloud12 - it is great you are wanting to get into open source contributions. Please apply the following principles to help protect the time of maintainers:
 - Check correctness of PR descriptions. The PR numbers in the summary of changes have nothing to do with the changes in this PR
 - For AI generated code, please make sure you have reviewed the code, and are confident that it is correct - and follows the libraries conventions
 - Make incremental changes, one issue per PR
 - Do not duplicate open work; there are already PRs open for some of the code changes here
 
A good piece of work that needs doing is updating the corresponding types packages in DefinitelyTyped - as there is quite a substantive drift between the types and the code at the moment. Remember that incremental changes are good; so you do not need to do the whole alignment in one PR.
 
 For this PR - please make sure the above principles are applied, then I would be happy to review. I have marked it as a draft until then.

PR #719, opened 2026-09-01
fix(parser): handle N3 verb edge cases

jeswr:
## Summary

- keep the N3 `has`, `is`, and `of` verbs from preempting longer valid prefixed names, including across stream chunks
- scope inverse-predicate state to the active predicate-object list: preserve it across commas and reset it across semicolons and annotation boundaries
- apply inverse direction consistently when closing blank-node property lists and formulas, and when constructing triple terms and reified triples
- add regression coverage for the cases identified in the [Copilot review of #694](https://github.com/rdfjs/N3.js/pull/694#pullrequestreview-5072997765)

This is a follow-up to #694.

## Verification

- 6,876 tests pass
- 100% statement, branch, function, and line coverage
- ESLint passes
- Node, IIFE browser, and ESM browser builds pass

### DIFF
--- a/src/N3Lexer.js
+++ b/src/N3Lexer.js
@@ -326,7 +326,7 @@ export default class N3Lexer {
       case 'h':
       case 'o':
         // Try to find an N3 verb keyword
-        if (this._n3Mode && (match = this._n3Verb.exec(input)))
+        if (this._n3Mode && (match = this._matchN3Verb(input, inputFinished)))
           type = match[0];
         else
           inconclusive = true;
@@ -336,7 +336,7 @@ export default class N3Lexer {
         // Try to find an IRI property list identifier or N3 verb keyword
         if (this._n3Mode && (match = this._n3Id.exec(input)))
           type = 'id';
-        else if (this._n3Mode && (match = this._n3Verb.exec(input)))
+        else if (this._n3Mode && (match = this._matchN3Verb(input, inputFinished)))
           type = match[0];
         else
           inconclusive = true;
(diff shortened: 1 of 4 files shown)


jeswr:
Performance follow-up for the inverse-handling concern:

I benchmarked `main` (`ffb71f9`) against this branch with Node 24.19.0. The harness alternated base/candidate order, used four fresh-process pairs with 20 samples per process, and included A/A (`main` versus `main`) controls. Parser workloads streamed 30,000 quads per sample; the isolated `_emit` workload created 2,000,000 quads per sample.

The initial implementation routed every `_emit` through `_createQuad` and showed a 5–8% paired-median slowdown in isolated emission. Commit `99d1e11` now installs the direction-aware `_createQuad` and `_emit` implementations only on N3 parser instances. For non-N3 formats, ordinary `_emit` remains the same direct factory call as `main`; blank-node and formula tails use a four-argument alias to that same method. This centralizes inverse swapping without putting inverse checks or a fifth argument on non-N3 emission paths.

Latest results (negative values are faster):

| Workload | Aggregate median | Paired median | A/A paired control |
| --- | ---: | ---: | ---: |
| Isolated `_emit` | -1.9% | -0.7% | +0.1% |
| Turtle | -4.3% | -2.9% | +4.9% |
| N-Triples | +0.7% | +3.4% | +2.3% |
| Turtle with blank-node property lists | -4.8% | -4.7% | +3.6% |
| Turtle with RDF 1.2 triple terms | -0.1% | -2.2% | — |

There is no measured non-N3 regression beyond the process/JIT/GC noise visible in the A/A controls. The only positive aggregate result is N-Triples at +0.7%; its paired result is 1.1 percentage points above its +2.3% A/A control, while its ordinary emission code is source-identical to `main`.

Verification on the revised branch: 6,876 tests pass with 100% statement/branch/function/line coverage; ESLint and the Node, IIFE, and ESM builds pass.

PR #729, opened 2026-09-05
perf(lexer): avoid match arrays for fixed tokens

jeswr:
Recognize `a`, boolean literals, and N3 `id` with `RegExp.test()` instead of `exec()`. Their matched text and length are already determined by the lexer branch, so a match array is unnecessary. The existing regexes keep the same token boundaries, incomplete-input handling, and line-mode restrictions; explicit lengths preserve token consumption and ranges.

This is independent of #727's separator dispatch and #728's literal-delimiter changes. It retains native regex scanning for the grammar while avoiding allocations in these fixed-token cases.

Direct string alternatives were also compared with this PR at `c5a91c1`: full `startsWith`/boundary-membership checks, direct words with a boundary regex, and space-only shortcuts. They preserve the tested behavior but did not establish a parser-level improvement for `a` or booleans. The seven-round confirmation below uses the same methodology as the original comparison, with medians and bootstrap 95% intervals:

| Alternative | Workload | CPU change | 95% interval |
| --- | --- | ---: | ---: |
| Full string matching | parser: abbreviation | +2.0% | [-19.0%, +28.9%] |
| Full string matching | parser: boolean | +8.4% | [-3.9%, +14.6%] |
| Full string matching | parser: keyword-like prefixed names | +5.8% | [+3.5%, +8.9%] |
| Space shortcuts for booleans and `a` | lexer: boolean | -5.8% | [-11.9%, -2.5%] |
| Space shortcuts for booleans and `a` | parser: boolean | +6.1% | [-4.1%, +30.7%] |
| Space shortcuts for booleans and `a` | parser: keyword-like prefixed names | +15.1% | [+5.0%, +22.2%] |
| Space shortcut for `a` only | parser: abbreviation | +18.9% | [-2.9%, +23.2%] |
| Space shortcut for `a` only | parser: keyword-like prefixed names | +13.4% | [+0.3%, +34.9%] |

The timings are noisy, and inconclusive differences should not be treated as gains or regressions. The full-string candidate and both space shortcuts each passed the applicable exhaustive boundary/streaming comparisons or retained an already checked path. Their prefix-heavy regressions do not justify replacing the current `.test()` calls. The native regex checks the word and its following boundary together; splitting that into extra string operations is not automatically faster. The prototypes are not included in this PR.

Seven paired rounds compare production Babel builds with main `e4e2148` on macOS arm64 / Node 25.1.0, rotating fresh-process order with 36 warmup and 36 measured iterations over 8,000-triple documents. Values are median paired CPU changes with bootstrap 95% intervals; negative means less CPU. Full token and parsed-quad digests agree.

| Workload | CPU change | 95% interval |
| --- | ---: | ---: |
| lexer: abbreviation | -16.5% | [-26.1%, -6.7%] |
| lexer: boolean | -5.9% | [-43.3%, -3.1%] |
| lexer: prefixed | -1.3% | [-3.5%, +9.5%] |
| lexer: dense | +1.7% | [-0.3%, +2.3%] |
| parser: abbreviation | -2.8% | [-7.5%, -2.4%] |
| parser: boolean | -3.8% | [-8.3%, +0.3%] |
| parser: prefixed | -0.7% | [-1.8%, +2.1%] |
| parser: dense | -4.6% | [-10.1%, +2.2%] |

The `a`-heavy parser workload improves in this run; the boolean parser and unaffected controls are inconclusive. These synthetic results do not guarantee the same gains for every application or runtime.

Validation: all **6,923 tests pass with 100% statement, branch, function, and line coverage**, plus ESLint and Node/browser IIFE/ESM builds. Tests verify exact token ranges and every stream split for `a`, `true`, `false`, and `id`. Additional assertions cover delayed emission until lookahead arrives, a following colon continuing a prefixed name, missing boundaries at EOF, line-mode rejection, and behavior with N3 features disabled. Differential checks against main cover **1,048,576 cases** (each following UTF-16 code unit in both line/N3 modes) and **1,242 stream splits**, preserving complete tokens, errors, and callback timing. Independent review found no correctness or maintainability blockers.

Rebased on main `73b38b3`, including the merged #727 separator dispatch and #728 literal delimiters. The two PR commits retain their reviewed changes. The rebased tree passes all 6,923 tests at 100% coverage, lint, and Node/browser builds. Benchmark comparisons above remain against their explicitly stated original bases.


src/N3Lexer.js
```diff
@@ -311,16 +311,16 @@ export default class N3Lexer {
       case 'f':
       case 't':
         // Try to match a boolean
-        if (match = this._boolean.exec(input))
-          type = 'literal', value = match[0], prefix = xsd.boolean;
+        if (this._boolean.test(input))
```
jeswr: Is it more performance to replace this with string operations

jeswr: Is it more performance to replace this with string operations

src/N3Lexer.js
```diff
@@ -311,16 +311,16 @@ export default class N3Lexer {
...
+          type = 'literal', value = firstChar === 't' ? 'true' : 'false', prefix = xsd.boolean, matchLength = value.length;
         else
           inconclusive = true;
         break;
 
       case 'a':
         // Try to find an abbreviated predicate
-        if (match = this._shortPredicates.exec(input))
-          type = 'abbreviation', value = 'a';
+        if (this._shortPredicates.test(input))
```
jeswr: Similarly, is it more perfromant to replace this with string operations

